PR #34, opened 2015-02-02
VS warnings

StDH:
Fixed warnings in Visual Studio compilers


StDH:
everything ok ?

texus:
Seems like you forgot the endif
https://travis-ci.org/texus/TGUI/builds/49178768

StDH:
yeah, i was curious why is it showing failed =D sorry.
This is the first time i use git.

texus:
It might be a good idea to learn git in the future so that you would be able to rebase these commits (so that the commits like "tabs to spaces" are part of the earlier commits and don't have a separate commit). But I don't really care that much about how the commit history looks, its just nicer to have less of these redundant commit. So its ok.

Thanks

PR #35, opened 2015-02-02
Changed ambiguity in some list constructors.

Ophilius:
Added overwrite function to signal::connect so you can have callback functions with 0 arguments.

### DIFF
--- a/include/TGUI/Gui.hpp
+++ b/include/TGUI/Gui.hpp
@@ -452,8 +452,8 @@ namespace tgui
         bool m_accessToWindow;
 
         // Internal container to store all widgets
-        GuiContainer::Ptr m_container = std::make_shared<GuiContainer>();
-
+        GuiContainer::Ptr m_container;
+        
         std::shared_ptr<Tooltip> m_visibleTooltip = nullptr;
         sf::Time m_tooltipTime;
         bool m_tooltipPossible = false;
--- a/include/TGUI/Signal.hpp
+++ b/include/TGUI/Signal.hpp
@@ -353,6 +353,34 @@ namespace tgui
         /// @return Id of this connection, which you need if you want to disconnect it later
         ///
         /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+        template <typename Func>
+        unsigned int connect(const std::string& signalNames, Func func) {
+           for (auto& signalName : extractSignalNames(signalNames))
+           {
+              if (m_signals.find(toLower(signalName)) == m_signals.end())
+                 throw Exception{ "Cannot connect to unknown signal '" + signalName + "'." };
+
+              try {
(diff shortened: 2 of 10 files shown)


texus:
I'm a bit hesitant to pull these changes. On one hand I don't have anything against them, they should work on all compilers. But on the other hand I can't reproduce your problems.

I just finished updating to the same visual studio version as you have, and I can still build tgui fine. So I am clueless on why you have compile errors while I don't.

The changes that were made right before you created this pull requests were even written in visual studio 2013, so I know at least one other person with that compiler that can build tgui fine.

I was kinda hoping that the difference would just be in the different visual studio version, that would have explained it, but it is not. So I have no idea why you can't compile it without those changes, but it looks like a problem on your end to me.

Ophilius:
How are you running visual studios? Are you running a virtual machine on your linux or is windows installed on the primary drive and what version of windows are you using to run it?

Currently, I am running on x64 Windows 8.1 pro with x32 visual studios 2013 ultimate and nvidia 780ti as the graphics card. I think those are all the software and hardware that would affect my compiling.

Here are the two bugs I am getting from the compiler. There are more C1001 hence the ambiguity fix:

C2783 Gui.hpp line 455: Could not deduce template argument for '_Ty' 
fix: I placed the creation of the object pointer in the three constructors.

C1001 listbox.cpp line 215: An internal error has occurred in the compiler
fix: I placed the casting for the list constructors

texus:
I'm on a 64-bit windows 7 with x32 visual studios 2013 professional. A real installation on a pc, no virtualbox. Graphic card is unrelated to this.

So those specs differ in operating system and VS version and in both cases I would expect yours to be better if there would be a difference at all.

When running CMake for the first time it will print the compiler version (delete all files in your build folder to get this again). Does it print MSVC 18.0.31101.0 for you too?

@TheIllusionistMirage @StDH Do you have Win8 and/or VS2013 ultimate? Then we can rule out these things as possible reasons.

Ophilius:
cmake reports me having MSVC 18.0.21005.1 RTM

texus:
Then that might be the cause.

In your mail you mentioned version 4.5.5, but that seems to be the .NET version. In the about screen in VS I have "Version 12.0.31101.00 Update 4". What does it say for you?

Ophilius:
Ok so it was my outdated MSVC. I updated MSVC from their website and it works fine now. However, I still have the problem with signal::connect not happy with callback functions with 0 arguments.

texus:
What is the exact message you get? And could you show how you are calling the connect function?

Ophilius:
Let me isolate the error. Right now I have about 80 errors, so it might be something one of the others. And thank you for helping. I greatly appreciate it.

Ophilius:
Updating MSVC++ fixed everything. Yay.

PR #38, opened 2015-04-02
Custom loader classes

Creepsky:
Hello :)

At first I have to say thank you for your effort programming such a nice and flexible gui library. You can be proud of yourself ;)

I dont know if I can contribute, I didnt find any rules, so I will just try it.

I work with the 0.7-dev branch and everythings works fine and neat. But I miss one thing: loading theme-files and images from memory.
Because it would be a loooot of work adding a new "create" function for every widget, I tried something else.
I made a virtual Loader-class for the TextureManager and for the ThemeFileParser. TextureManager and ThemeFileParser have both static loader-instances, both shared_ptr, so they can be polymorphic.

The idea behind is: you can inherit from the Loader-classes and override the standard load-function.
The default loader-classes implements the load-function like they were implemented before.

You can set the static Loader-instances by overriding them with the static functions "setFileLoader" for the ThemeFileParser and "setImageLoader" for the TextureManager.

For example you can combine the load function with a custom resourcemanager or just load it from memory.

If there is a better option, please tell me :)


texus:
Thanks.

I am planning to completely rewrite the way widgets are loaded in v0.7, so eventually your changes will be thrown away though.

But since this probably won't happen before the end of July or so, I will accept this pull request so that others can make use of your changes.

PR #39, opened 2015-06-01
Add vertical and horizontal layout containers.

Dank-Y:


texus:
A few remarks on the code:
- Copy constructors, assignment operators and BoxLayout constructor are not needed
- Using a Button as a space widget is a serious overkill. Perhaps use a ClickableWidget instead
- All tabs should be converted to spaces

Are you going to fix these or should I do it?

Dank-Y:
I'll erase copy constructors /assignment operators, I just forgot to do it.
I used button because that allow the user to set a custom texture to the space, but as you said ClickableWidget is maybe better. If the user want put an image as separator, it is better to let him do it himself, just like for other widgets.
I'll replace tabs.

texus:
Thanks a lot.
I'll still make some minor changes to it to get the code exactly how I want it and then I'll merge it later today.

Dank-Y:
As you seen in the build, would it not be better to use std::size_t or size_t instead of unsigned int for indexes ? (I think that was also done in the last version of SFML, btw).

texus:
The build on AppVeyor with all these warnings about size_t and unsigned int is actually the very first time that TGUI was build as a 64-bit library. So I'm not really surprised that I get so many warnings.

I will eventually fix all these warnings by using size_t I guess, just not now.

texus:
I had already merged your branch locally and added a similar add/insertSpace function.
I'll just manually copy code from your changes and add them to what I have now.

texus:
Does the space really still need a name? Since the other functions are now based on the widget pointer or an index.

texus:
I've made the commit of the fixed size under your name manually.
If you still have an idea to improve these classes then you should get rid of your current changes and use the latest version of my repo because the branches are no longer compatible.

I have already pushed all the changes but they still require some testing.

Dank-Y:
The identifier for spaces was useful when spaces where Buttons, but they are effectively useless now.

Dank-Y:
Actually I think the fixed size do not work, you have made the needed changes in BoxLayout.* but not in Vertical/HorizontalLayout.cpp (mainly the updateWidgetPositions method).

texus:
I didn't test fixed size yet, I'll look into it.

PR #40, opened 2015-06-02
V0.7 dev

Dank-Y:
Checkboxes and Sliders behave wrong when they are using inside a layout.
Theses changes are not the best way to fix that, but that's the idea.

### DIFF
--- a/include/TGUI/BoxLayout.hpp
+++ b/include/TGUI/BoxLayout.hpp
@@ -204,7 +204,7 @@ namespace tgui
         /// @return False when the index was too high
         ///
         /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-        bool setFixedSize(unsigned int index, float ratio);
+        bool setFixedSize(unsigned int index, float size);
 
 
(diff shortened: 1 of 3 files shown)


texus:
They are not the only widgets that are broken and this is definitely not a solution since you just hardcode some values. The problem is that widgets are bigger than the size that you give them (due to borders or like the thumb in slider).

Just to be sure that we are talking about the same kind of issues (I haven't checked the checkbox or slider yet), removing glScissor in Panel.cpp on line 163 solves the issue, right?

Dank-Y:
For sliders that solve the issue, but checkboxes scales themselves the hard way and when putted into a layout, the square and the V in the checkbox fit the whole cell, I don't think that is the right behavior when making a form for example.

texus:
The size of a checkbox is very tricky, it has to be the size of the box.

The real solution to this layout problem is adding a setFullSize function to widget. Just like getFullSize returns the size of the checkbox plus the size of the text, the setFullSize would calculate the optimal sizes for both box and text. Layouts would then use setFullSize instead of setSize. But this is a pretty big change.

PR #59, opened 2016-03-16
Fix typo and format of BoxLayout documentation

TheoVerhelst:
This PR fixes some spelling mistakes I made when I wrote this class. And also removes dots that shouldn't be there.

I also can do this work for your other classes, have a clean and consistent documentation is more comfortable for the reader.

If I do so, can you confirm that you use these conventions for the doc?
* Only long explanations end with a dot, other fields (`@brief`, `@param`) don't;
* Methods brief doc are written in the third person;
* Class briefs end with a dot;

### DIFF
--- a/include/TGUI/BoxLayout.hpp
+++ b/include/TGUI/BoxLayout.hpp
@@ -52,7 +52,7 @@ namespace tgui
 
 
         /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-        /// @brief Changes the size of the layout.
+        /// @brief Changes the size of the layout
         ///
         /// @param size  The new size of the layout
         ///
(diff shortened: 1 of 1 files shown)


texus:
There isn't really a convention yet, the style has changed a few times during development which is why it sometimes lacks consistency.
I'm not sure where to put a dot myself. On one hand I would like to drop it everywhere, but then there will always be inconsistencies with long explanations where dots are required between sentences and where I find it normal to end with a dot.

But it may not be a bad idea to define a convention and try to stick to it from now on:
- I guess for the dots, just remove them everywhere at the end of `@param`, `@brief` etc. When e.g. `@param` would consists of multiple sentences, the last sentence would not have a dot.
- Sentences that are not part of one of these fields and are thus there to give more information about the function should always end with a dot.
- I typically write in third person so that is something to keep

Feel free to do it for other classes as well, but for the sake of preventing huge merge conflicts, could you do documentation changes in `0.7-WIP-beta-rewrite` instead of `master`?

texus:
I have manually merged it into 0.7-WIP-beta-rewrite in 907390e

TheoVerhelst:
Thanks, I'll do similar work for the rest of the documentation as soon as possible.
Your proposal for dot convention seems good, it is like the one used by [SFML](http://www.sfml-dev.org/style.php#comments-documentation).

PR #60, opened 2016-03-19
More consistent documentation

TheoVerhelst:
As we discussed earlier, I updated the documentation to respect the following rules:
* The last line of all fields starting by `@` has no dot;
* The verb of all `@brief` of functions is conjugated in the third person;

**Note:** there is conflicts, I tried to resolve them but in some files I don't understand them, so I have to let you resolve them yourself.


texus:
Apparently you didn't update the branch since last week which causes some merge conflict. But I'll fix them myself. The changes look great but I'll have a better look at them later.

Could you merge your commits into one though?

TheoVerhelst:
I merged the commits. Sorry for the conflicts, I fetched the branch '0.7-WIP-beta-rewrite' today, I guess I did something wrong in the process.

texus:
If by fetching you mean running "git fetch" then that is not enough. It will only download the changes from my repo but you would still have to merge it with "git merge". It should also be possible with just a "git pull" though. I always have to look it up again when I need to do stuff like this.

texus:
The changes have been merged. Thanks a lot for your contribution.

TheoVerhelst:
You're welcome!

PR #62, opened 2016-07-07
ChildWindow Resizing Fix

Ruckamongus:
With the recent additions to the ChildWindow I neglected to test resizing. Currently, you can resize a window with more than one button and the extra button(s) will hang off of the edge. This PR fixes this by only allowing the ChildWindow to be resized as small as the sum of the width of the buttons and their padding.

I also changed the way buttonOffsetX was determined (instead of each button hardcoded it loops more elegantly like your previous changes).

See here (red borders for dragging bounds):
image

Unrelated, if I get time I may send another PR allowing the upper part of the window to resize (as it is now the title bar cannot be used to resize). This doesn't seem like a quick addition though, and some design changes may be required: currently the draggable area is defined by the renderer borders. This works okay, but I like to draw a single crisp pixel around the window and clicking a single pixel to drag is very difficult. What I propose is keep the borders strictly for drawing and having another function, something such as setResizeWidth(std::size_t pixels), that allows us to split the drawing borders and the draggable area borders. Furthermore, there may be instances that being able to resize in only one plane desirable. Something such as setResizePlanes(Resize::Vertical) would only allow vertical resizing. What are your thoughts on this before I start playing around?

Edit: On second thought I'm not sure that the plane thing is necessary. I thought I had a good use in mind for it but it doesn't seem like something that's all that useful.

### DIFF
--- a/src/TGUI/Widgets/ChildWindow.cpp
+++ b/src/TGUI/Widgets/ChildWindow.cpp
@@ -89,12 +89,12 @@ namespace tgui
 
         // Calculate the distance from the right side that the buttons will need
         float buttonOffsetX = 0;
-        if (m_closeButton->isVisible())
-            buttonOffsetX += (buttonOffsetX > 0 ? getRenderer()->m_paddingBetweenButtons : 0) + m_closeButton->getSize().x;
-        if (m_minimizeButton->isVisible())
-            buttonOffsetX += (buttonOffsetX > 0 ? getRenderer()->m_paddingBetweenButtons : 0) + m_minimizeButton->getSize().x;
-        if (m_maximizeButton->isVisible())
-            buttonOffsetX += (buttonOffsetX > 0 ? getRenderer()->m_paddingBetweenButtons : 0) + m_maximizeButton->getSize().x;
+        for (const auto& button : {m_closeButton, m_maximizeButton, m_minimizeButton})
+        {
+            if (button->isVisible())
+                buttonOffsetX += (buttonOffsetX > 0 ? getRenderer()->m_paddingBetweenButtons : 0) + button->getSize().x;
+        }
+
         if (buttonOffsetX > 0)
             buttonOffsetX += getRenderer()->m_distanceToSide;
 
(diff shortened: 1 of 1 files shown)


texus:
In the mouseMoved I would call the variable something like minimumWidth instead of buttonOffsetX.
I don't see how the code is working correctly on first sight, I would think buttonOffsetX is missing the distance between the last button and the side of the child window. I would expect that you needed m_distanceToSide twice. But I'll do some testing tonight or tomorrow.

The draggable area being defined as the rendered borders was only a quick hack to get resizable windows without much effort, it was never intended to be permanent. The way I was thinking of doing it is simply use the most outer pixels (the ones that now serve as a border if you set the size to 1, but also the ones in the title bar) plus 3 or 4 more pixel rows/columns which lie outside of the window as draggable area. I think this can remain hard coded for now, I don't think you need a setResizeWidth function, it will only make things more complicated.

Having a function for only horizontal or vertical resizing is less useful. The ability to set a minimum and maximum size might be more interesting and can achieve the same effect by e.g. setting the minimum width and maximum width to the same value.

I'm still waiting for [this pull request](https://github.com/SFML/SFML/pull/827) to make the resizing look nice though. When that pull request finally ends up in SFML then there should be some way in which TGUI widgets can change the cursor type (e.g. resize icon in child window) but there should of course also be a way for the user to override this behavior and not let TGUI change the icon.

texus:
The offset in mouseMoved is incorrect as I suspected (it just isn't really visible unless you e.g. set distanceToSide to 20)

The following lines:
```c++
if (buttonOffsetX > 0)
    buttonOffsetX += getRenderer()->m_distanceToSide;
```

Should be replaced with:
```c++
minimumWidth += 2 * getRenderer()->m_distanceToSide - getRenderer()->getBorders().left - getRenderer()->getBorders().right;
```

Ruckamongus:
> In the mouseMoved I would call the variable something like minimumWidth instead of buttonOffsetX.
That's a good point, I'll change that.

> The way I was thinking of doing it is simply use the most outer pixels...
That sound really good! I'd like to contribute, but I really don't know how to do that. How do you insure there is no other widget or window under the mouse? I might just hold off and hope the fellas at SFML decide on something (it's unfortunate how slow things seem to go over there. Understandable mostly but it still sucks sometimes).

> The following lines: ... Should be replaced with: ...
I'll make the changes tonight. Do I need to squash the changes again or should I hold off for more testing?

texus:
> How do you insure there is no other widget or window under the mouse?
You can assume there isn't one. By enabling resizing you will allow the window to make use of the few pixels around it as well. This is the reason why the resizeable property should remain false by default.
Right now it might become weird because the mouse cursor won't change and people might expect the click to go to the widget behind it, but once there are mouse cursors it will become clear that when the mouse is at that location the click will go to ChildWindow.

Keep in mind that the mouseOnWidget function should also return true when on top of these invisible pixels around the child window when the window is resizeable.

> Do I need to squash the changes again or should I hold off for more testing?
I don't think I need to do any more testing so you can immediately squash the commits. You actually don't even have to create a new commit at all, just commit using the "--amend" flag. Your changes will be added to the existing commit (they are added to the last commit you made). You will still have to push with "--force" because the commit online and locally will no longer be the same.

Ruckamongus:
Thanks for the git help!

I wonder how certain cases will work. Consider this image (red is the resize bounds):
image


If my mouse is on one of the blue spots I'd expect to be resizing the above window. If it was on the green spot it would technically be in the "Below" window's resize bounds. How do I handle these scenarios? 

In the GUI I wrote for the game I'm porting I didn't have outer resize boundaries, only inside. I then just resized the contents panel inwards to account for the resize border. It did take some playing around, especially for the titlebar buttons, but it worked off of knowing the mouse was in the current window. Something like this:
image

texus:
> How do I handle these scenarios?

They are already handled by TGUI. On either location both the mouseOnWidget functions will return true, but only the top one will receive the events. So mouseOnWidget will be called on both windows, but mouseMoved is only called on the top one. So the case will work as expected without needing to write any extra code.

With having the resize area on the inside you might get other issues, e.g. what if the area is larger than distanceToSide? Part of the close button won't be clickable as it will be considered as a resize.

I don't really care whether the resize area is inside, outside or a combination of both, as long as it work a bit intuitively.

Ruckamongus:
Finally got around to this, sorry for the delay. I'm not sure how I should handle the resizing if you're wanting the outside pixels to work too (which I agree would be nice); mouseMoved() will not be called if the mouse is outside the boundaries so I'm not sure how to go about it yet.

texus:
> I'm not sure how I should handle the resizing if you're wanting the outside pixels to work too (which I agree would be nice); mouseMoved() will not be called if the mouse is outside the boundaries so I'm not sure how to go about it yet.

The mouseOnWidget function should also return true when on top of these invisible pixels around the child window when the window is resizeable.
Basically when the mouse is on top of this invisible area (and the resizeable flag is true) then the same thing should happen as when the mouse is on top of the title bar (see first if case in mouseOnWidget).

PR #63, opened 2016-08-18
Update for SFML 2.4

Cmdu76:
I used SFML_VERSION_MAJOR and SFML_VERSION_MINOR to ensure that the library will be usable for SFML 2.4 and older versions.

### DIFF
--- a/src/TGUI/Loading/ThemeLoader.cpp
+++ b/src/TGUI/Loading/ThemeLoader.cpp
@@ -34,7 +34,7 @@
 #include <fstream>
 
 #ifdef SFML_SYSTEM_ANDROID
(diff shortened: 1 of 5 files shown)


texus:
Thanks for contributing.

There are two reasons why I can't pull this though:
- The code won't work for a version like 3.0, the conditional if check should be "major > 2 or (major == 2 and minor >= 4)"
- You used tabs instead of spaces in some lines

Could you perhaps patch these things?

Cmdu76:
Just a question about my commits and newer commits, are they automatically applied to the build downloadable on the website with travis and appveyor ? Or I have to build it myself ?

texus:
You have to build the github version yourself, the downloads on the website are not updated.

Cmdu76:
Thank you for your answer and being so reactive !

PR #64, opened 2016-09-11
Allow the ChildWindow to have maximum and minimum sizes.

Ruckamongus:
Alright, so this allows the user to explicitly set a ChildWindow's minimum and maximum sizes when resizing by dragging the edges. It's easiest to test by modifying the ChildWindow theme properties to something like:
```
ChildWindow {
    BorderColor     : rgb(255, 0, 0);
    Borders         : (4, 4, 4, 4);
}
```

Some behavioral decisions:
- setSize() does not respect any previously set maximum/minimum sizes
- Resizing width will stop to keep the buttons within the titlebar even if the set minimum size is smaller. For example, a ChildWindow with 2 buttons may require a minimum of 50 pixels to keep the buttons on the titlebar properly. If you set the minimum width of the ChildWindow to be 30, the window will not be resized below 50.
- Negative values are allowed and not checked for. Per @texus: "I consider negative size undefined behavior in widgets"

The main.cpp code for testing is a bit sloppy, but on [pastebin](http://pastebin.com/raw/YE5m2tpr) for convenience.

### DIFF
--- a/include/TGUI/Widgets/ChildWindow.hpp
+++ b/include/TGUI/Widgets/ChildWindow.hpp
@@ -152,6 +152,52 @@ namespace tgui
 
 
         /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+        /// @brief Returns the maximum size of the child window.
+        ///
+        /// @return Maximum size of the child window
+        ///
+        /// The size returned by this function is the maximum size of the child window, excluding the title bar and the borders.
+        ///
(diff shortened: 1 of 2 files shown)


Ruckamongus:
I forgot about the constexpr thing and C++11's restraints. Is TGUI meant to support C++11? If so I should change that clamp() implementation.

texus:
Merged in e717b44dae50cb8800e116368f661791dd162f4c with some changes.

PR #67, opened 2017-02-02
Added stdio include

Scellow:
I wasn't able to compile it using MinGW 

```
 error: 'EOF' was not declared in this scope
             while (stream.peek() != EOF)
```


texus:
Thanks, apparently testing on 5 different compilers wan't enough to get standard compliant code 😄

Could you change it to <cstdio> though? Then I'll merge it.

Scellow:
Hmm i'm not sure if I understood what you mean ^^, what do you want me to change?

texus:
I forgot one word in that sentence :)
It should have been "Could you change it to <cstdio> though?"
Because stdio.h is the c library instead of the c++ one. It shouldn't make any difference, it is just to be consistent with other includes.

Scellow:
Ohh, right! i'll do it now

Scellow:
Done :)

PR #69, opened 2017-05-28
Added a Container that flows onto additional lines.

cyanskies:
I wrote this for use in my editor, but someone else might find it useful. It's a modification of HorizontalLayout that moves extra widgets onto a second line rather than scaling them to make them fit.

It's useful for listing many similarly sized buttons, like the "toolbox" windows in image editors.

I use it for displaying an arbitrary set of tiles in a ChildWindow in my editor.


texus:
Thanks.
I'll have a look at this when I rewrite the HorizontalLayout and VerticalLayout in a month or two.

texus:
Merged in 0.8-dev (4cc813c5ef5a2e9de81c3b208bb40932d54d07c4).

Issue #72, opened 2017-08-24
Static TGUI requires Static SFML


Nopey:
I'm about to make a PR to remove `TGUI_SHARED_LIBS` triggering `SFML_STATIC` and `SFML_STATIC_LIBRARIES` flags.

IMO the `SFML_STATIC` & `SFML_STATIC_LIBRARIES` flags' existence should be announced when  `TGUI_SHARED_LIBS` is false, but not automatically set.

Nopey:
Good compromise, Thanks!

PR #73, opened 2017-08-24
Static TGUI no longer automatically uses SFML Static Libs

Nopey:
See Issue #72 for more info
TODO: Consider announcing the SFML Static libs flags when compiling static TGUI without static SFML.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -270,6 +270,11 @@ if(NOT TGUI_SHARED_LIBS)
     endif()
 endif()
 
+# Warning if they have set TGUI to static libs, but not set SFML to static.
+if((NOT TGUI_SHARED_LIBS) AND (NOT SFML_STATIC) AND (NOT TGUI_QUIET_STATIC))
+  message(WARNING "TGUI_SHARED_LIBS is false, but SFML_STATIC is false. TGUI will link statically, but SFML will be dynamically linked. Pass -DSFML_STATIC to cmake to link SFML statically. Pass -DTGUI_QUIET_STATIC to disable this message")
+endif()
+
 # Find sfml (also look for the main component when using Visual Studio)
 if (SFML_OS_WINDOWS AND SFML_COMPILER_MSVC)
     find_package(SFML 2 COMPONENTS main graphics window system)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,8 +238,6 @@ endif()
 
 # Link to the static sfml libraries when building tgui statically
 if(NOT TGUI_SHARED_LIBS)
-    set(SFML_STATIC_LIBRARIES TRUE)
-
     # Attempt to find the SFML dependencies when linking statically
     if (SFML_ROOT)
         if (SFML_OS_WINDOWS)
@@ -321,12 +319,6 @@ endif()
 # We need to link to an extra library on android (to use the asset manager)
 if(SFML_OS_ANDROID)
     set(TGUI_EXT_LIBS ${TGUI_EXT_LIBS} "-landroid")
-endif()
-
-# Add SFML_STATIC define when linking statically and link to SFML dependencies
-if(NOT TGUI_SHARED_LIBS)
-    add_definitions(-DSFML_STATIC)
-    set(TGUI_EXT_LIBS ${TGUI_EXT_LIBS} ${SFML_DEPENDENCIES})
 endif()
 
 # Generate .gcno files when requested


texus:
I don't understand your second sentence in #72, could you clarify the effects of your change? Also, what is your use case? I don't see any reason why you would want to link SFML and TGUI in a separate way.

Won't your change always link the dynamic SFML libraries when TGUI is linked statically? My tests are failing on your change (https://ci.appveyor.com/project/texus/tgui/build/0.8-build-455/job/bj3lq66161jyf8le), probably because only static SFML libraries exist and it tries to find the dynamic ones.

TGUI currently also relies on SFML_STATIC when compiling your own program, so you can't link SFML and TGUI in a different way without changing this. I could make my own TGUI_STATIC define but that means users will have do make an extra define to get TGUI working statically and I see no reason to make it more complicated for beginning programmers.

Nopey:
The reason I want to statically link TGUI, and dynamically link SFML is that SFML is in most package manager's repositories, but TGUI probably isn't.
I know there are binaries for windows, (or were those dev binaries?).

what do you mean about TGUI currently relying on SFML_STATIC? I think you might be talking about the automatic flagging of SFML_STATIC and SFML_STATIC_LIBRARIES in your cmakelists.txt, which is what this PR removes.

The test is failing because the test has static SFML available, but isn't passing in -DSFML_STATIC.

I think that the change between having to call
`cmake /path/to/tgui -DTGUI_SHARED_LIBS=FALSE`
to
`cmake /path/to/tgui -DTGUI_SHARED_LIBS=FALSE -DSFML_STATIC`
is pretty easy for beginner programmers with a warning to them when they have TGUI_SHARED_LIBS=FALSE, but not SFML_STATIC. This is what my second sentence meant.

I'm about to add the warning, to let developers using the flag the old way know why their builds suddenly start being unable to find their static SFML.

texus:
For windows there are binaries, but only if you use the corresponding SFML version.
What package manager where you using that contains SFML but not TGUI (I know about a few, but it is interesting to know which are actually being used so that it may be worth publishing binaries there as well)?

TGUI relies on SFML_STATIC inside [Config.hpp](https://github.com/texus/TGUI/blob/0.8-dev/include/TGUI/Config.hpp) as well. With the current contents of that file, you cannot link TGUI differently than SFML on windows without getting issues.

Even with the warning I'm not happy about this change. I believe almost everyone will link SFML and TGUI in the same way and they should not have to make an extra define just to make it work. On windows, the define even has to be made in 2 places: once in cmake when compiling TGUI and once in your own project (where the contents of Config.hpp will play a role).
So I would accept a pull request that allows linking SFML and TGUI differently somehow, but not when the user has to perform extra steps by default just to get static linking to work.

Nopey:
I'm using Ubuntu 17.04, and there is no TGUI package in the default repository

texus:
For ubuntu I have a PPA that allows easily installing TGUI though, which I guess most people will use that instead of building TGUI statically.
https://launchpad.net/~texus/+archive/ubuntu/tgui-0.7

Nopey:
My reasoning behind statically linking to TGUI but not SFML is my code will probably work with whatever SFML the user has installed, but TGUI is going under some big exciting changes, and I'd like my program to continue working even if I don't update and fix my program to work with each new stable shared library version's API tweaks.
And while I could distribute a TGUI binary with my program, thats the problem that static linking is supposed to solve.

texus:
I was actually going to write a list of extra changes which should be made before I would accept this pull request, but I think the use case may not be valid based on your last post.

> reasoning behind statically linking to TGUI but not SFML is my code will probably work with whatever SFML the user has installed

This depends on the perspective. This may be true if recompiling your application on each pc it runs on but if you ship an executable with TGUI statically linked to a different pc which has a different SFML version installed I can almost guarantee you that it won't work.

Neither SFML nor TGUI has ABI stability between versions, so you can't update the version of the library without recompiling your program anyway (which is why you should always ship the dependencies with your executable or ship just the source code and the build system).

The API of TGUI 0.7 is stable, the one in 0.8-dev is indeed changed regularly. But this is only relevant for compiling the application again. SFML and TGUI are c++ libraries where it is difficult to keep the same ABI and there is no effort to keep it stable. If you e.g. update SFML on your pc then you must recompile some programs, this has happened multiple times to me already. You may not notice this when using ubuntu where all library versions stay stable, but on an arch linux where you regularly update to the latest versions of packages, I have to occasionally recompile my programs after SFML or GCC updated.

If you really only care about specific ubuntu versions and you will be providing an executable for each of these ubuntu versions then you might have a valid case, as the ppa can still be updated while the sfml version can't, which may break the program when dynamically linking. But if you care about any other linux distribution or you want to create a single executable that runs on multiple ubuntu versions then you are going to need a different approach and can't rely on SFML being installed in the package manager.

eXpl0it3r:
This PR in itself is certainly not enough for a good CMake experience, but thinking of supporting different mixes of linking is certainly something that should be discussed.

Once you set SFML_STATIC_LIBRARIES the FindSFML.cmake will automatically define SFML_STATIC.

texus:
For cmake I guess it could be implemented like this:
``` cmake
if (NOT DEFINED SFML_STATIC_LIBRARIES)
  if (TGUI_SHARED_LIBS)
    set(SFML_STATIC_LIBRARIES FALSE)
  else()
    set(SFML_STATIC_LIBRARIES TRUE)
  endif()
endif()
```

The user could set SFML_STATIC_LIBRARIES if he really wants to specify how SFML is linked, but if not set, SFML will be linked the same way as TGUI.

The Config.hpp issue however also has to be solved. Since the use case that I see for it is extremely small, I'd make it harder to mix the linking to keep it more simple for the other 99% of the users.
There should be an extra flag (I'm not sure what a good name could be, e.g. TGUI_MIXED_LINKING) which has to be set as well. If not set, the linking stays as it is now. Only when the flag is set will TGUI not rely on SFML_STATIC to determine the linking.

The top of the Config.hpp could be changed to something like this then (with the defines with lowercase characters being placeholders for a potential better name):

``` c++
#ifdef tgui_mixed_linking
    // TGUI will be linked statically when TGUI_STATIC is defined, SFML_STATIC only affects SFML linking
    #define tgui_static_internal TGUI_STATIC
#else
    // TGUI will link in the same way as SFML, no extra define is needed for TGUI
    #define tgui_static_internal SFML_STATIC
#endif
```

The `#ifndef SFML_STATIC` below should then become `#ifndef tgui_static_internal`

Edit:
In CMakeLists.txt, the `add_definitions(-DSFML_STATIC)` should be removed as this will be handled in FindSFML.cmake as eXpl0it3r said, but the `set(TGUI_EXT_LIBS ${TGUI_EXT_LIBS} ${SFML_DEPENDENCIES})` should remain. The line can actually always be executed, independently of TGUI_SHARED_LIBS as SFML_DEPENDENCIES should be empty when linking dynamically.

PR #86, opened 2018-12-31
Ability to disable Word Wrap on the TextBox widget

xandayn:
This PR adds the ability to disable word wrapping on the TextBox by adding a horizontal scrollbar to it. The horizontal scrollbar being enabled disables word wrap, and disabling it enables word wrap.

Also adds a utility function in the Text class to measure the width of a string (single line) and adds a few extra filters to the .gitignore for building directly in Visual Studio.


texus:
Thanks for the contribution. Could you use spaces instead of tabs everywhere though? I can't review your changes because every line in TextBox was changed.

texus:
I've pulled the code and replaced the tabs locally. It seems like the 2 hacks should still be sorted out.

Did you test the performance of Text::getLineWidth? I've looked at the implementation and unless I'm missing something it should be extremely fast to do this for all lines, it should even be faster than executing the word-wrap (which can handle HUGE texts). So if the call would only be executed when the horizontal scrollbar is visible, then you either run wordWrap or getLineWidth on all lines but never both. That way you wouldn't lose performance when not using the horizontal scrollbar and you might even gain performance when using a horizontal scrollbar (as getLineWidth should be faster than wordWrap).

Could you perhaps explain the hack in updateSelectionTexts?

xandayn:
The hack in updateSelectionTexts moves the m_caretPosition calculation out of recalculatePositions to before we attempt to move the scrollbar based on said caret position. I had issues (again due to monospace vs non-monospace fonts) in properly calculating how far the scrollbar should move, so I fell back on using m_caretPosition to help with the calculation, but when you paste some text, the scrollbar would move before the new caret position would be calculated, so I had to move it. I don't believe it changes any functionality of the original TextBox, but I left the comment just in case there was something I overlooked.

xandayn:
On the question about the performance of getLineWidth, I don't think the function itself has a huge issue with performance, I think that if I put the m_lines calculations into the if/else block of !isHorizontalScrollPresent() I could iterate over all the lines that way, I still don't know for sure if it would be optimal or not, but it's definitely an option, and would solve the issue of monospace/normal fonts.

texus:
I prefer to have a solution that works well for all fonts.
I just tested the performance of Text::getLineWidth, it only takes 0.04 seconds to get the length of a 1000 lines of 100 characters each. So we can get the real length of all lines without having to worry about performance and we don't have to rely on hacks that only work for monospace fonts.

Instead of moving the m_lines calculations into the !isHorizontalScrollPresent() if/else block (which would probably lead to a little code duplication in the if and else blocks), maybe you can just replace
```c++
if (m_lines.back().getSize() > maxStrCount)
{
    maxStrCount = m_lines.back().getSize();
    longestLine = m_lines.size() - 1;
}
```
with something like:
```c++
if (isHorizontalScrollbarPresent())
{
    float lineWidth = Text::getLineWidth(m_lines.back(), m_fontCached, m_textSize);
    if (lineWidth > m_maxLineWidth)
        m_maxLineWidth = lineWidth;
}
```

The m_maxLineWidth can remain undefined (or set to 0 somewhere) when the horizontal scrollbar isn't visible. The variable is also used in updateScrollbars, but the case where it is used isn't needed when there is no horizontal scrollbar (`m_horizontalScroll->getSize().y` never has to be subtracted from the vertical scrollbar size if there can't be a horizontal scrollbar). So with a little restructuring it should be possible to only access m_maxLineWidth when the horizontal scrollbar is enabled.

I'll try to actually test your changes tomorrow. Based on what you said I guess I can just remove the comment and keep the changes in updateSelectionTexts.

xandayn:
In my own tests, I've found it takes about ~0.01s per ~140,000 characters (~600 lines, ~200 characters per line with a line of about ~650 every ~30 lines or so) with my current method and ~0.25s for the same data using the per line method to calculate the maximum line width (only if the scrollbar is visible). It's quite the extreme increase. Although it's not the biggest performance hit, that appears to be coming someone later down the line (rendering is my guess) which causes the real slowdowns. I don't have proper testing set up to find where the real bottleneck is.

texus:
I get very different results than you. It takes ~0.02s for 140,000 characters (which I assumed was because my laptop with an i5-7200U isn't as fast as your pc), but when calculating all lines it takes only ~0.06s in debug mode and ~0.045s in release mode. I measured the duration of just the rearrangeText when typing text, as this is more relevant than the duration of the setText call which will be executed less often (the difference with the duration of setText wasn't much though except during the first setText call where rearrangeText is executed twice). If you randomly mash multiple keys on the keyboard then there is a visible delay before the letters appear on the screen, but if you type normally then 0.045s per typed character is still fast enough.

If you disable the horizontal scrollbar then everything is even slower, so even if the per line calculation takes a long time it is still an improvement over what existed before.

You don't have to make any more changes to this PR, I'll implement the changes that I still want myself. Since I don't want code that works only for monospaced text, I will make the more expensive calculation the default (most people won't use such huge texts anyway). I will however add an `enableMonospacedFontOptimization(bool)` function so that you still have to choice to use the optimized version.

At the end of the rearrangeText function you call rearrangeText again if `(horiScrollbarShown != m_horizontalScroll->isShown())`. The original code was there because if the vertical scrollbar became visible then the word-wrap needed to be re-executed. If the horizontal scrollbar becomes visible, then you don't really need to recalculate anything about the text. Am I correct that I can remove that change again and only call rearrangeText there when the vertical scrollbar changed?

texus:
Merged in a97296eda75f025524e004d70d9183c73604ebe6

xandayn:
It's possible that the recursive call at the end doesn't need to be called again, I honestly didn't test it very much. I'm not even sure why I changed it in the first place to be perfectly honest. I probably had a reason but I can't remember. Also I agree that it shouldn't only work properly for monospace fonts, but I just couldn't think of a solution I was happy with to do it otherwise.

PR #95, opened 2019-03-24
GuiBuilder: Widget hierarchy

PeacefulPieDev:
изображение

Added a widgets hierarchy window.


texus:
Thanks for contributing.

I tested it and it seems like renaming a widget is broken if that widget isn't added directly to the form. If I create a panel with a picture inside it and rename the picture then the name in the hierarchy window doesn't change and clicking on the picture again shows the old name back in the properties window.

The border of the widget hierarchy is also too thick because both the child window and the tree view have borders. The tree view should be defined in EditingScreen.txt (instead of created in c++) with "borders = (0, 0, 0, 0)" in the renderer, while the ScrollablePanel.Widgets section in HierarchyWindow should be removed (as it isn't used).

Except for these small issues, the hierarchy window looks good.

If you are going to implement saving/restoring the child window positions then you should watch out what you save exactly. The windows can currently be manually dragged, but after dragging them they will no longer keep their relative position when the main window is resized and you can never get them back like they were when the program started. So if their position is saved then you will never be able to get them back in the right spot. Maybe we shouldn't allow the windows to be dragged (i.e. set `PositionLocked = true` for child windows)? Or maybe the windows should somehow be rearranged when dragging stops (i.e. not allow them to be placed anywhere on the screen)?

I'm willing to overlook code style issues and manually fix them when merging, but it would be easier for me if the code would use spaces instead of tabs and if statements with only one line inside them wouldn't contain braces around that line.

texus:
My last change introduced a small conflict in the code as I changed the createNewWidget function. You basically just need to have the new createNewWidget code plus the "widgetHierarchyChanged();" line that you added there.

The other conflicts are inside the recursiveCopyWidget and copyWidget functions which have been removed. These functions weren't actually used anymore. The functions that are being used to copy widgets call createNewWidget so they don't need further changes and will already update the hierarchy.

texus:
I've merged your changes in e4c401383741aa70ca5d31bacb0d455e05002885
You can open a new PR if you make further changes (but make sure to first sync with my latest changes).

PR #100, opened 2019-07-29
Added compile features to CMake build script

Yanson32:
When I link against TGUI with CMake I have to set the c++ standard to 14 manually. Adding a list of compile features will make that  unnecessary..

### DIFF
--- a/src/TGUI/CMakeLists.txt
+++ b/src/TGUI/CMakeLists.txt
@@ -146,6 +146,19 @@ else()
     set_target_properties(tgui PROPERTIES RELWITHDEBINFO_POSTFIX -s)
 endif()
 
+# required compiler features
+target_compile_features(tgui PUBLIC
+                                cxx_nullptr
+                                cxx_auto_type
+                                cxx_override
+                                cxx_right_angle_brackets
+                                cxx_strong_enums
+                             PRIVATE
+                                cxx_decltype
+                                cxx_range_for
+                                cxx_lambdas
+)
+
 tgui_set_global_compile_flags(tgui)
 tgui_set_stdlib(tgui)
 


texus:
I got some mixed feelings for this. On one hand I think it is a good idea to do this, but on the other hand I'm not really a big fan of target_compile_features.

Maybe things have gotten better by now, but when I last looked into it there were some issues with target_compile_features.
- Some c++17 features would only be available when using a recent CMake version (and I try to keep the minimum supported version low enough to compile TGUI on linux distros that are a few years old). Setting the CXX_STANDARD standard property to 17 had much better support I think.
- Also some compilers wouldn't be recognised. A quick search made me find [this](https://stackoverflow.com/a/28136384/3161376), but I'm not sure how relevant it still is. The comments on the answer seem to suggest that there still is some issue on mac though.

I feel like I have more control when setting it manually as opposed to letting CMake pick the option (and it would still be an equal amount of work for me to check which features I'm allowed to use in my code without forcing the latest compiler to everyone). In the private 0.9 branch that I created I e.g. added a cmake option that you could set to 14, 17 or 20. I would put the value in a define in Config.hpp so that the code would compile based on the supported c++ version. Of course this design is far from ideal, it would be much better to rely on feature testing and enable feature by feature based on what the compiler supports. I actually did that where possible (e.g. using char8_t if available) but it just didn't work well enough yet and I still needed to manually enable things (I think it had something to do with __has_include not being usable).

I don't have that much experience with cmake and I'm not sure how to combine this conditional compilation with target_compile_features. What features should I ask for in cmake, the minimum ones (and force non-optimal code to be compiled even though your compiler supports better) or the maximum ones (and force you to compile using the highest possible c++ standard even though you have other parts of the code that require a lower standard)?

In the end, unless you can provide some suggestions to deal with these concerns, I feel like I should keep it as is because this might solve one issue but potentially create some others.

Maybe instead of writing this wall of text I should have asked a different question as well.
In which cases is it actually necessary to set it to c++14 manually in your own project? Visual Studio, GCC >= 6 and clang >= 6 use c++14 by default. The discussion for whether to add this pull request are still valid for future versions, but I guess there aren't many people who are actually going to notice that the c++ standard is currently not specified when using TGUI with cmake?

texus:
Based on the comments on [this CMake issue](https://gitlab.kitware.com/cmake/cmake/issues/17225), it seems like the `target_compile_features` function with individual features isn't the way forward. Instead the meta-features such as `cxx_std_14` and `cxx_std_17` (which were added in CMake 3.8) are recommended. These meta-features correspond better to what I want.

`target_compile_features(${target} PUBLIC cxx_std_14)` should be able to replace `set_target_properties(${target} PROPERTIES CXX_STANDARD 14)` while having the advantage that it also gets applied to your project.

Instead of merging this PR I've made my own commit that uses target_compile_features with cxx_std_14, but thanks for letting me know that I could use target_compile_features to make using TGUI with a cmake project easier. If you wouldn't have created this PR then I wouldn't have looked for a way to improve it.

So this issue should be fixed, at least when using CMake 3.8 or higher.

PR #102, opened 2019-08-11
Gui Builder: saving form size to EditorState.txt

PeacefulPieDev:


texus:
The change itself is good, but could you perhaps still change 2 things?
- There are some tabs instead of spaces
- I'm not fond of the parseLayout being made public for this. I feel like it is overkill to use code that can load complex layout strings just for parsing 2 integers. Maybe this could be simplified? (You could e.g. just copy the parseLayout function but remove most of it, all that you need to check are the brackets at begin/end and then just split on the only comma).

PeacefulPieDev:
Ok, I will review tabs and apply your suggestion to simplify parsing size.

PeacefulPieDev:
Updated the PR according to proposed changes.

PR #104, opened 2019-08-22
Ability to toggle mouse wheel changing Slider value

Miv99:
Works exactly like ComboBox's changeItemOnScroll, but with Slider instead.


texus:
I overlooked this PR, I didn't knew it existed until just now (I must have been busy when I got the email notification and forgot to check it afterwards). Thanks for contributing.

PR #107, opened 2019-09-07
Add sort for ListView

Kvaz1r:
I not think that it's the best solution but as for me it's still better than removing all items and adding them back on each sorting. 

If these changes is appropriate I'll also add tests.


texus:
This definitely looks like a useful function.

Could you perhaps swap the parameters of the sort function (so first the index and then the comparator)? I think that it might be slightly more readably when calling the function with a lambda function as comparator.

I don't really like that both variables are initialized on one line, so maybe you could also either use a ternary operator (and make s1 and s2 const) or rewrite the function as follows?
```c++
sf::String s1;
if (index < a.texts.size())
    s1 = a.texts[index].getString();

sf::String s2;
if (index < b.texts.size())
    s2 = b.texts[index].getString();
```

Kvaz1r:
good points, I'll fix it.

PR #108, opened 2019-09-16
allow TGUI to be build simultaneously with SFML

bebu:
The idea is following:

i have following git repo:

SFML (submodule to 2.5.x branch)
TGUI (submodule to 0.8 branch)
awesome_game (actual source code)
CMakeLists.txt

this patch allows sth like this in CMakeLists.txt:
```
add_subdirectory(SFML)
add_subdirectory(TGUI)
include_directories(
   ${SFML_SOURCE_DIR}/include
   ${TGUI_SOURCE_DIR}/include
)
add_subdirectory(awesome_game)
```

Perhaps, instead of autodetection, one could use an option

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,31 +206,28 @@ if(NOT TGUI_SHARED_LIBS)
     endif()
 endif()
 
+# Find sfml if not build in same project
+if(NOT TARGET sfml-graphics)
+    if(TGUI_OS_WINDOWS AND TGUI_COMPILER_MSVC) # Also look for the main component when using Visual Studio
+        find_package(SFML 2 COMPONENTS main graphics window system)
+    elseif(TGUI_OS_ANDROID)  # Search for SFML in the android NDK (if no other directory is specified)
+        find_package(SFML 2 COMPONENTS graphics window system PATHS "${CMAKE_ANDROID_NDK}/sources/third_party/sfml/lib/${CMAKE_ANDROID_ARCH_ABI}/cmake/SFML")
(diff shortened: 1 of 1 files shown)


texus:
If it works then I'm fine with adding something like this.

If you could make the following changes then I will merge this PR:
- The message isn't very descriptive, but instead of improving it, I would just remove it. Just have an if statement that checks that the target does NOT exist and don't have an else case. A comment could be added right above the if statement to say that the SFML libraries are being searched for unless TGUI and SFML are being build in the same project.
- `sfml-graphics` should probably be checked instead of `sfml-window`. It doesn't matter much, but it is sfml-graphics that is being used in target_link_libraries so it would fit better.
- Why is the LIBRARY_OUTPUT_PATH variable also part of this if statement? I would move that line out of the if branch, unless you have a good reason for it.

bebu:
I adopted your ideas.
The idea behind LIBRARY_OUTPUT_PATH was, that the parent project does set this already. But i think thats a bit too much tailored to my case :), so moved the endif before that.

texus:
Maybe setting the LIBRARY_OUTPUT_PATH variable can be changed to something like this?
```
if(NOT DEFINED LIBRARY_OUTPUT_PATH)
    set(LIBRARY_OUTPUT_PATH "${PROJECT_BINARY_DIR}/lib")
endif()
```

If that works then you can still easily overwrite it by setting LIBRARY_OUTPUT_PATH in your own project while people who don't define it themselves still have the same behaviour as before.

bebu:
Actually it is not needed as it seems, setting it in parent scope seems to be enought, so the way it is, its okay for me :)

PR #110, opened 2019-09-28
Update Canvas.hpp

Unicode-Hater:
Added a "cast" function to RenderTexture for Canvas.
What's the idea?
Let's just say that you have a function for drawing thing in your program that takes a reference of RenderTarget to draw all things thus perfect for passing a window, a RenderTarget or a RenderTexture. I saw that the Canvas is a wrapper of RenderTexture, so I thought that I can add a "cast" function that can be used in that particular scenario.


texus:
Thanks for making the PR.
I ended up implementing this myself instead of merging this PR because I wanted a few changes (different description and using reference instead of pointer as value can't be a nullptr), and for such a small commit it was easier to just implement it myself compared to explaining everything and asking you to do it.
So a getRenderTexture function returning a reference to the render texture has been added in 9c941efc4342acce36bdd8df2b774f8ad65afe06.

PR #111, opened 2019-09-29
Add getItemColumn for ListView

Kvaz1r:
There is already function getItem/0 for extracting values from first column so I propose to add method for extracting values from special column.

### DIFF
--- a/include/TGUI/Widgets/ListView.hpp
+++ b/include/TGUI/Widgets/ListView.hpp
@@ -381,6 +381,16 @@ namespace tgui
 
 
         /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+        /// @brief Retrieves the value for a cell in the list
+        ///
+        /// @param rowIndex The index of the row
+        /// @param columnIndex The index of the column
+        ///
+        /// @return Texts of the cell or an empty vector when the index is out of range
+        /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+        sf::String getItemCell(std::size_t rowIndex, std::size_t columnIndex) const;
+        
+        /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
         /// @brief Returns a list of the texts in the first column for all items in the list view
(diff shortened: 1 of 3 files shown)


texus:
In what cases would such a function be useful?
I prefer to know at least one use case of how a function would be used before adding it.

Kvaz1r:
For example if listview provide a table with scores for one set of user and I have two sets with scores for two different periods and I want to show who is increase his result and who is decrease. User can exists in one set and not exist in other so I extract from specific column the list which consist all  username that I want to use.

texus:
The name isn't very clear to me, maybe getColumnValues would be better?
I tried to look up what it would be called in other libraries, but the ListView from Windows doesn't seem to have such functionality.

Wouldn't you need to loop over the list that getItemColumn returns? Because it doesn't really add extra functionality in that case compared to what can already be done:
```c++
for (unsigned int i = 0; i < listView->getItemCount(); ++i)
    doSomethinng(listView->getItemRow()[2]);
```
```c++
const std::vector<String>& values = listView->getItemColumn(2);
for (const auto& value : values)
    doSomething(value);
```

An argument could be made that the version with getItemRow would be slower because it has to return the entire row just to access a single value from it, but that would be solvable by adding a function to retrieve a specific cell in the ListView.

Kvaz1r:
> The name isn't very clear to me, maybe getColumnValues would be better?

Yes, naming isn't my cup of tea. 

> An argument could be made that the version with getItemRow would be slower because it has to return the entire row just to access a single value from it, but that would be solvable by adding a function to retrieve a specific cell in the ListView.

It's exactly the argument. Function to retrieve specific cell will be universal so it's obvious the best solution. I can update the PR with it or create a new one tomorrow.

texus:
You can just update this PR.
I guess calling it `getItemCell(rowIndex, columnIndex)` should be clear enough (I'm not good at making up names either).
The current getItem function should continue to exist, but would just call `getItemCell(index, 0)` internally.

PR #113, opened 2019-10-11
Add multiselect to ListView

Kvaz1r:
I decided to add setSelectedItems function too because I don't know how to provide tests without it.  
I am not a native English speaker so there are could be grammar errors in descriptions.


texus:
Thanks. The code looks fine on first sight. I'll merge it soon, after I had the time to look at the code in more detail.

texus:
On closer inspection I found that the code wasn't polished enough to be merged as-is. It contained several issues such as text colors not being updated properly and not being able to unselect items except by clicking on another item (which wasn't possible once all items were selected).
I ended up rewriting several parts of the code. You might want to check out the new code and test that everything still works and verify that I didn't accidentally remove something that you added. Function names are a bit different as well, I ended up using "Indices" instead of "Indexes" (both are correct English, but .NET also uses "Indices" for their ListView as well).

PR #126, opened 2020-02-29
Update license to 2020

Xrey274:


texus:
Thanks. This has to happen on more places though, so I've pushed a commit that changes it everywhere.

PR #127, opened 2020-04-12
Update hscroll maximum taking into account vscroll width

Kvaz1r:
Without these changes horizontal scroll will appear together with vertical scroll. 
MCVE:
https://gist.github.com/Kvaz1r/1534bab797b62e7fd8faf78b4eacb581

### DIFF
--- a/src/TGUI/Widgets/ListView.cpp
+++ b/src/TGUI/Widgets/ListView.cpp
@@ -901,7 +901,7 @@ namespace tgui
             m_verticalScrollbar->setAutoHide(true);
         }
 
-        updateScrollbars();
+        updateHorizontalScrollbarMaximum();
     }
 
     /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
@@ -932,7 +932,7 @@ namespace tgui
             m_horizontalScrollbar->setAutoHide(true);
         }
 
-        updateScrollbars();
+        updateHorizontalScrollbarMaximum();
     }
 
     /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
@@ -1900,7 +1900,7 @@ namespace tgui
             maximum += static_cast<unsigned int>((m_items.size() - 1) * m_gridLinesWidth);
 
         m_verticalScrollbar->setMaximum(maximum);
-        updateScrollbars();
+        updateHorizontalScrollbarMaximum();
 
         // If the scrollbar was at the bottom then keep it at the bottom
         if (verticalScrollbarAtBottom && (m_verticalScrollbar->getValue() + m_verticalScrollbar->getViewportSize() < m_verticalScrollbar->getMaximum()))
@@ -1926,7 +1926,10 @@ namespace tgui
             else
                 width += m_columns.size() * static_cast<float>(getTotalSeparatorWidth());
 
-            m_horizontalScrollbar->setMaximum(static_cast<unsigned int>(width));
+            if (m_verticalScrollbar->isVisible())
+                m_horizontalScrollbar->setMaximum(static_cast<unsigned int>(width - m_verticalScrollbar->getSize().x));
+            else
+                m_horizontalScrollbar->setMaximum(static_cast<unsigned int>(width));
         }
 
         updateScrollbars();


texus:
I think it should only take the vertical scrollbar into account when there is one. If the vertical scrollbar isn't visible then it should still have the full width as maximum.

Kvaz1r:
Hmm,  if vertical bar is just unshown it still have non-zero size, for example at creation.

texus:
Yeah, that's why the size should only be subtracted when its isVisible() returns true.
This also means that on places where the vertical scrollbar might become visible (e.g. when the maximum value of the vertical scrollbar is changed), then the maximum of the horizontal scrollbar has to be updated again (I haven't looked at the code to see where changes might be needed).

Kvaz1r:
> This also means that on places where the vertical scrollbar might become visible (e.g. when the maximum value of the vertical scrollbar is changed), then the maximum of the horizontal scrollbar has to be updated again (I haven't looked at the code to see where changes might be needed).

I'd say it could happen every time when vertical scroll update own maximum. Should I update PR in this way or there are cases when it not true?

texus:
There seems to be only one place where the maximum is changed (updateVerticalScrollbarMaximum), but it looks like you also need to update the horizontal scrollbar in updateScrollbars (since the size of the vertical scrollbar is changed there).
Those two places should be enough, you can update the PR with that.

Kvaz1r:
> but it looks like you also need to update the horizontal scrollbar in updateScrollbars (since the size of the vertical scrollbar is changed there).

updateScrollbars is [called from updateHorizontalScrollbarMaximum()](https://github.com/texus/TGUI/blob/0.8/src/TGUI/Widgets/ListView.cpp#L1932) so it will be recursion. But updateScrollbars() is only used in 4 places so I replaced where it appropriate.

texus:
We're going to have to come up with a better solution for this. Your code first updates the maximum of the horizontal scrollbar and only then calls updateScrollbars which might cause the vertical scrollbar to change which in turn might require the horizontal scrollbar to be updated (but the maximum isn't updated again after updateScrollbars is executed).

This is basically the same problem as I faced in ScrollablePanel, the scrollbars depend on each other (horizontal scrollbar could be invisible if vertical scrollbar is hidden but whether vertical scrollbar is visible might depend on whether the horizontal scrollbar is shown). But I don't really like my code in ScrollablePanel::updateScrollbars, I was hoping there was an easier way.

If you want to then you can try to come up with a proper solution to the circular dependency, but otherwise I will have a look at this later.

texus:
Also, I just realized something, why does the maximum need to change? The maximum depends on the contents of the ListView, the ScrollablePanel::updateScrollbars doesn't set the maximum depending on whether the other scrollbar is visible. So you might even be looking at a fix in the wrong place, the maximum should not depend on the scrollbar sizes.

This error is happening because you set the sizes of the columns to perfectly match the size of the ListView without scrollbar. If you make the last column a variable size then you wouldn't have an issue with the horizontal scrollbar.

Kvaz1r:
> So you might even be looking at a fix in the wrong place, the maximum should not depend on the scrollbar sizes.

Here my reasoning. 
1. Maximum for hscroll is set initially and won't change until size of column not changed.
2. Scroll will be shown while maximum is more that size. 
3. When vscroll is became visible (or vice versa) it's updated size for hscroll inside updateScrollbars. Difference is m_verticalScrollbar->getSize().x. 

Oh, maybe just need to add update for maximum at those conditions. 

> This error is happening because you set the sizes of the columns to perfectly match the size of the ListView without scrollbar. If you make the last column a variable size then you wouldn't have an issue with the horizontal scrollbar.

I know it, I remember corresponding forum [thread](https://forum.tgui.eu/help-requests/how-to-create-listview-with-width-equal-of-window-and-without-redundant-scroll/) :)

texus:
The ListView is doing exactly what you are asking from it. You are creating one column of 160px and one column of 640px. If the scrollbar takes 20px then it can only display 780px out of the 800px of contents, so it displays a horizontal scrollbar.

The problem is that the ListView has no way of having a column that fills 80% of the available size with the available size becoming smaller when a vertical scrollbar is shown. This is what you probably want, but this is currently unsupported.
You would have to subtract the scrollbar width from the `clientWidth` in your code, or fall back to the dynamically sized last column.

Based on the functionality that exists right now, the ListView is acting incorrectly. If you subtract the vertical scrollbar width from the maximum of the horizontal scrollbar then you will always be missing the last 20px (or whatever the scrollbar width is) in the last column.

Kvaz1r:
Ok, I completely confused here and can't see any good solution.  So close the PR as there is simple workaround. 


> You would have to subtract the scrollbar width from the `clientWidth` in your code, or fall back to the dynamically sized last column.

I don't see corresponding function in the documentation otherwise I would already did so :)
getScrollbarWidth() from renderer can return only non-default value. I set it to 16 but it would be better also to have opportunity get default value.

texus:
I still don't know how to best handle customization with widgets that contain a scrollbar.
I added a getScrollbarWidth function to ScrollablePanel to get access to that information, so the short-term solution would probably be to copy that function to ListView as well.

PR #132, opened 2020-06-25
Recalculate tabs width after removing

Kvaz1r:
For some reason remove is not calling recalculateTabsWidth(), opposite for all others modify operations.


texus:
Seems like an oversight. Thanks.

PR #134, opened 2020-07-03
Fix work with position for SubwidgetContainer

Kvaz1r:
In every method Container already make position shifting([example](https://github.com/texus/TGUI/blob/0.9-dev/src/TGUI/Container.cpp#L659)) so it's no need to do it twice. 

Also after [this](https://github.com/texus/TGUI/commit/91748a182ab9ace65883392e2024a41d86c336f7#diff-1cf979e4b47d461afa0485bef59c8e65) commit Container translates position so accordingly update this widget too.


texus:
Does this fix anything? Your commit message seems to suggest so, but the code looks as if you just moved some code from one place to another?
I would still merge it even if it doesn't fix anything just because it makes the code slightly cleaner, but it would have to be with a different commit message if no problems existed before this change.

Kvaz1r:
> Does this fix anything? Your commit message seems to suggest so, but the code looks as if you just moved some code from one place to another?

Yes, MCVE (it should set red background color for spin buttons on hover). Before the fix, code didn't work as expected (even at 0.8 branch) and in 0.9 there is also appears small drawing issue:

P2


code(for 0.8 mouseOnWidget instead of isMouseOnWidget):
```
#include <TGUI/TGUI.hpp>

class SpinCtrl : public tgui::SubwidgetContainer
{
public:
    typedef std::shared_ptr<SpinCtrl> Ptr; //!< Shared widget pointer
    typedef std::shared_ptr<const SpinCtrl> ConstPtr; //!< Shared constant widget pointer

    SpinCtrl(const tgui::Layout2d& size = { "100%", "100%" })
    {
        m_type = "SpinCtrl";
        setSize(size);

        m_spinButton = tgui::SpinButton::create();
        m_spinText = tgui::EditBox::create();

        m_spinButton->setPosition(tgui::bindRight(m_spinText), tgui::bindTop(m_spinText));

        m_spinText->setSize(m_spinText->getSize().x, m_spinButton->getSize().y);
        m_spinButton->getSharedRenderer()->setBackgroundColorHover(tgui::Color(255, 0, 0));

        m_container->add(m_spinText);
        m_container->add(m_spinButton);
    }

    static SpinCtrl::Ptr create(const tgui::Layout2d& size = { "100%", "100%" })
    {
        return std::make_shared<SpinCtrl>(size);
    }

    static SpinCtrl::Ptr copy(SpinCtrl::ConstPtr spinctrl)
    {
        if (spinctrl)
            return std::static_pointer_cast<SpinCtrl>(spinctrl->clone());
        else
            return nullptr;
    }

    Widget::Ptr clone() const override
    {
        return std::make_shared<SpinCtrl>(*this);
    }

    bool isMouseOnWidget(tgui::Vector2f pos) const
    {        
        return m_container->isMouseOnWidget(pos);
    }

private:
    tgui::SpinButton::Ptr m_spinButton;
    tgui::EditBox::Ptr m_spinText;
};

class MyFrame
{
public:
    MyFrame()
    {
        window.create(sf::VideoMode(800, 600), "0.9 After");
        gui.setTarget(window);

        auto button = tgui::Button::create("button");
        gui.add(button);

        auto spin = SpinCtrl::create();
        spin->setPosition(tgui::bindRight(button), tgui::bindBottom(button));
        gui.add(spin);
    }

    void main()
    {
        while (window.isOpen())
        {
            sf::Event event;
            while (window.pollEvent(event))
            {
                    window.close();
                gui.handleEvent(event);
            }
            window.clear(sf::Color(200,240,220));
            gui.draw();
            window.display();
        }
    }
    sf::RenderWindow window;
    tgui::Gui gui;
};

int main()
{
    MyFrame().main();
}
```

texus:
The SubwidgetContainer was never really tested. I was planning to implement a SpinEdit using it to demonstrate its usage, but I never found the time for it so the code was included without being used anywhere.

The change to `SubwidgetContainer::draw` is definitely a bug introduced in 0.9.

Whether the other position change classifies as a bugfix is debatable. The isMouseOnWidget function was responsible for subtracting the position, just like all other functions also did this. The real "bug" to me is that you had to override isMouseOnWidget yourself. I think there should be a default implementation in SubwidgetContainer so that you only have to override it when your widget isn't rectangular. Your change is obviously still an improvement, whether I call it a bugfix or not.

Could you add the following to your fix as well (and declare the function in the header)?
```c++
bool SubwidgetContainer::isMouseOnWidget(tgui::Vector2f pos) const
{        
    return m_container->isMouseOnWidget(pos);
}
```

I would also change the commit message to something like "Fixed issues with setting position of SubwidgetContainer", but I could do that myself when merging.

Thanks for your work and for the MCVE and images, they were very helpful, especially since I currently don't have a functioning 0.9 version on my pc and I thus couldn't easily test it here.

Kvaz1r:
> Could you add the following to your fix as well (and declare the function in the header)?

Yes, updated.

PR #137, opened 2020-07-18
Update spin value periodically if left mouse pressed is been holding

Kvaz1r:
I'm not sure that Timer::scheduleCallback is the right choice here, but idea is simple - move changing value from mouse release handler to mouse pressed handler and to call changer periodically while some conditions (mouse still pressed, mouse still hover on widget and correct spin button and there wasn't any new clicks) are still being true.

### DIFF
--- a/include/TGUI/Widgets/SpinButton.hpp
+++ b/include/TGUI/Widgets/SpinButton.hpp
@@ -29,6 +29,7 @@
 
 #include <TGUI/Renderers/SpinButtonRenderer.hpp>
 #include <TGUI/Widgets/ClickableWidget.hpp>
+#include <TGUI/Timer.hpp>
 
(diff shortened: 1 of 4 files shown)


texus:
I like the idea of using Timer::scheduleCallback.
Although it is unlikely to happen, I fear that it could cause issues if the widget is destroyed between scheduling the callback and receiving it. Instead of capturing `this` in the labmda, it would probably be better to capture something like `spinButton=shared_from_this()` and add `spinButton->` in front of variables and functions inside the lambda. That way the widget can't be destroyed until after the scheduled lambda has been executed.

Kvaz1r:
Fixed

texus:
Manually merged as 8e8db091984917c65b1ca3e56c1c8edbb4965a35

PR #138, opened 2020-07-24
Add insertItem function to ListView widget

AmbiSykhar:
Fairly self-explanatory, useful to have; without it, adding an element mid-list would require editing every entry after it.


texus:
Thanks for contributing. I do see 2 issues with the code though:
- You do not check whether the index is within the items. Something like the following should probably be added:
```c++
if (index >= m_items.size())
{
    addItem(...);
    return;
}
```

- The scrollbar will scroll to the bottom of the list view when m_autoScroll is true, no matter where the items are inserted. Ideally it would scroll to the inserted item, but since nobody has asked for such functionality, you don't have to implement it. I would also accept that you simply don't call setValue at all on the scrollbar, so basically just removing the following code from your functions:
```c++
// Scroll down when auto-scrolling is enabled
if (m_autoScroll && (m_verticalScrollbar->getViewportSize() < m_verticalScrollbar->getMaximum()))
    m_verticalScrollbar->setValue(m_verticalScrollbar->getMaximum() - m_verticalScrollbar->getViewportSize());
```

AmbiSykhar:
Ah, whoops, you're right. I thought I was missing something. Pushed a fix to both of those, and went ahead and added the autoscroll, since it's just fairly simple, unless I'm missing something else.

texus:
The tests are failing to build in c++17 mode. Apparently the vector::emplace function returns an iterator instead of a reference.
So you would need to dereference it so that `auto& object = vector.emplace(vector.begin() + index);` becomes `auto& object = *vector.emplace(vector.begin() + index);`

I wasn't going to mention it, but now that I'm asking for another change anyway, maybe you could also change `vector.at(index);` by `vector[index];`? The at() function usually isn't used in TGUI (there is no need for out-of-bound exceptions).

Then I will be able to merge the code.

AmbiSykhar:
I was wondering what the issue there was. Apologies, I've been working a little too much with C# lately, my C++ is a little rusty. Looks like emplace has returned an iterator since C++11, so I'm taking it out of the conditional as well, no need for it to be in there.

PR #141, opened 2020-08-17
Added possibility to sort widgets.

1aam2am1:
Added possibility to sort Container::m_widgets by std::sort and user lambda compare function to further enable moving widgets.

In example now can I sort layers of widgets and change what should be first displayed in one command and not by calling moveWidgetForward or moveWidgetBackward

### DIFF
--- a/include/TGUI/Container.hpp
+++ b/include/TGUI/Container.hpp
@@ -107,6 +107,19 @@ namespace tgui
             return m_widgets;
         }
 
+        /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+        /// @brief Sorts a list of all the widgets in this container
+        ///
+        /// @param function comparison function object (i.e. an object that satisfies the requirements of Compare) which
+        ///                 returns true if the first argument is less than (i.e. is ordered before) the second.
+        ///
+        /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+        template<typename Function>
+        void sortWidgets(Function&& function)
+        {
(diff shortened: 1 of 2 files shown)


texus:
What use case do you have for this?
Layers of widgets could be created by putting them into Groups, then you just have to move the group forward or backwards.

1aam2am1:
Yes but I can have many groups of widgets say 5. Now to create sixst and place it in say 3 place I need place it and move widget 3 times. 
But I don't know in what place I need to place my new group with example z-order 128. I need to place it then check all widgets (get name from them) and remember with z-order was will be and what to move up or down.(sort them) Or copy vector of widgets clear widgets then sort my vector and add it to gui one more time. 
In this way I can change order and sort my container based on different things.

texus:
I'm still not really convinced that a sort function is really the best way to handle this.

Based on the first part that you said, about creating a sixth group and moving it to place 3, I would suggest adding a function that moves a widget to an absolute position in the list (as opposed to requiting you to call a function multiple times that only moves it one spot each time). This function was considered but ultimately not added because I couldn't figure out whether the index that you pass to such function should be the index before or after the widget is removed from the list.

The index in the list is currently the z-order. I do understand that some people might want to use a custom z-order and let the container automatically order the widgets based on this custom number. I believe such an implementation would be superior to the sorting, because new widgets could be inserted in the right place instead of having to run a sort on all widgets again.

The sort function does provide full control over the order and thus provides a lot of flexibility, which the Container class currently lacks, so I will merge this. But the sort function might be removed again in the future if I ever implement a better way to specify a z-order (not very likely to happen soon, and now that the sort function exists it would have an even lower priority).

PR #142, opened 2020-08-22
Add include cassert

Kvaz1r:
Without including cassert I get error C3861 identifier not found (VS19 in release mode).


texus:
Thanks for reporting. I'm surprised that this issue wasn't caught by AppVeyor, the CI successfully builds TGUI in release mode in VS2019.

The assert calls were planned to be replaced by TGUI_ASSERT. Instead of fixing this issue by adding the missing include I have decided to just replace all remaining assert calls right now.

PR #143, opened 2020-08-23
GuiBuilder: Add SpinControl support

Kvaz1r:
Implementation quite straightforward because it's indeed simple. New change from other widgets is way to distinct properties for spin button from spin text's properties - I use set for each and prefix "Buttons" if text part has exactly same property. 

I don't know how to generate icon for resource (gui-builder\resources\widget-icons) so for local test I used copy file for spin button widget and not include it here.

Single remaining question is how to correctly handle "[change back property part](https://github.com/texus/TGUI/blob/0.9-dev/gui-builder/src/GuiBuilder.cpp#L660)" after updating property, but I'm not understand how to test it (for me all works even without this line).


texus:
For the image I just create something simple with GIMP. The following should do:
image

That line is responsible for e.g. setting the property to "#643232" when you type "rgb(100,50,50)" as value. For simple properties such as strings or numbers you won't see a difference with or without that line.

One potential solution would be to add another virtual function to WidgetProperties. The setProperty call should be moved to there. SpinControl could then override that function and call setProperty on the correct widget.

An alternative would be to add a rendererChanged function in SpinControl. All properties would e.g. be called either "Edit.xxx" or "Spin.xxx" (or "EditBox.xxx" and "SpinButton.xxx" or similar) and SpinControl::rendererChanged would then call renderer->setProperty on either the editbox or spinbutton renderer. This way you wouldn't need to m_buttonProperties and m_textProperties in SpinControlProperties and you would just be able to call WidgetProperties::updateProperty there (as long as you also prefix your renderer properties in initProperties).

Kvaz1r:
> For the image I just create something simple with GIMP. The following should do:

Cool, should I include your image here or you going to add it afterwards? 

> That line is responsible for e.g. setting the property to "#643232" when you type "rgb(100,50,50)" as value. For simple properties such as strings or numbers you won't see a difference with or without that line.

Yes, I see now. In that case it's enough to call again updateProperty because property grid already contains 100% correct value. Fixed.

texus:
You can include the image in this PR.

PR #147, opened 2020-09-24
Allow horizontal scrollbar for ListView items

Miv99:
These changes allow the horizontal scrollbar to appear to view items too long to fit in a ListView. 

When there are no columns, the scrollbar's max value will be enough to view up to the end of the longest item. 

When there are 1 or more columns and m_expandLastColumn is false, the ListView will follow old behavior.

When there are 1 or more columns and m_expandLastColumn is true, the scrollbar's max value will be enough to view up to the end of the longest item in the rightmost column. Items in other columns will be clipped by their column's width (like in old behavior).


texus:
Thanks for contributing. I'm a bit concerned with the potential performance impact of this change though. With optimizations the code you wrote would be welcome, but it won't be accepted in its current form.

I noticed you are calling updateColumnsMaxItemWidths every time you add a single item. This means that if you add 1000 lines, you need to calculate 500,000 line widths.

There are several optimizations that are probably possible (based on a quick look at the changes):
- For adding items, you only need to check if that one item is larger than the maximum.
- For removing items, if the width was smaller than the maximum you don't need to recalculate anything.
- The line widths could be cached so that you only need to calculate them once. If you do need to loop over all items (e.g. when removing the longest line) then you at least don't need to recalculate everything.
- Since there is no need for these calculations when there are columns and m_expandLastColumn is false, I don't think the max width should be calculated at all in such case.
- The maxWidth calculations add things such as textPadding and iconWidth on each item. It would probably be better to calculate the maximum line width and then only add the padding and icon width after you found the longest line. (This is a micro-optimization that I don't care much about though, I'm just including it because I'm listing all the other stuff anyway)

Miv99:
Thanks for the suggestions. I implemented bullet points 1, 2, and 4 in this commit but skipped the 3rd because I was having some trouble figuring out how to cleanly cache line widths in TextView. The caching could be done pretty easily in Text but would require making getLineWidth() no longer const and I'm not sure if that should be avoided.

texus:
One last performance thing to look into is maybe only calling updateHorizontalScrollbarMaximum when needed. It probably only needs to be called when the maximum width changes instead of for each item.

There is some duplicate code in changeItem and changeSubItem, maybe the lines starting at updatedLastColumnMaxItemWidth could be placed in a helper function (with oldDesiredWidthInLastColumn as parameter).

The first two lines in addMultipleItems seem to be leftovers from before, so they can probably be removed.

I looked into whether getLineWidth can be cached and I came to the conclusion that it might not be what you need. The function is only used in one place and was added as a helper function during some PR. You can probably just use `getSize().x` instead, which does cache its value internally.

The `Text::getLineWidth` without parameters doesn't even seem to be used anywhere in TGUI (and can thus be removed), while the only use case of the other getLineWidth function would be to calculate the width of the text without first having to create a Text object (and for that reason it might be better to move the function to the Font class). But that it something I'll look into later.

Miv99:
Implemented all those (except the Text things you're looking at) in this commit and fixed some edge case with my code for changeSubItem() with no columns.

texus:
Thanks for adding the requested optimizations. Everything looks good now.

PR #152, opened 2020-12-05
Add file dialog windows drives support

### LINKED ISSUE #151: File dialog don't support changing drives in Windows
It should be possible to switch in file dialog between drives.

Kvaz1r:
Fix: #151

### DIFF
--- a/src/Widgets/FileDialog.cpp
+++ b/src/Widgets/FileDialog.cpp
@@ -29,6 +29,10 @@
 #include <map>
 #include <ctime>
 
+#ifdef TGUI_SYSTEM_WINDOWS
+    #include <TGUI/WindowsInclude.hpp>
+#endif
+
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 namespace tgui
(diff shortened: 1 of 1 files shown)


texus:
Thanks. Functionality wise it looks good (I didn't bother to test it, but I understand what the code does and I trust you enough by now).

Could you make a few minor code changes though? (I only really care about the first two points, but since I'm asking to change those I might as well ask for the other changes as well)
- The windows.h include should be removed from the header. In the source file you can add the following. That will hopefully fix the build failure.
```c++
#ifdef TGUI_SYSTEM_WINDOWS
    #include <TGUI/WindowsInclude.hpp>
#endif
```
- The LogicalDrives variable should start with a lowercase character
- You don't need the i variable, addItem will return the index, so you can write `const std::size_t i = m_listView->addItem(...);`
- Is the `= { 0 }` needed? I think GetLogicalDriveStringsW will either have written the required the zeros or it would have returned false, so the array can probably remain uninitialized.
- Use std::wcslen instead of wcslen
- You don't need "::" in front of GetLogicalDriveStringsW

Kvaz1r:
Fixed comments from review and added hiding vertical gridlines and header when drives list shown.

Issue #164, opened 2021-12-30
Container get function don't search tab container panels


Groogy:
If I am trying to find a widget by name that is in a panel that belongs to a tab container it will not be found. Quick look at the code is that it will be skipped by `Container::get` since it is not a container itself.

Would love if I could search the content of the panels as well instead of having to fetch the tab container first and then search each tab independently.

texus:
I seem to have missed the notification about this, I'm only seeing this now.

This really is an edge case. The SubwidgetContainer class was meant for widgets that need a container internally but aren't containers themselves. TabContainer inherits from SubwidgetContainer but then adds multiple panels inside this internal container. Technically those panels are an implementation detail of TabContainer and aren't really part of the widget tree. But I can see why you would want the get function to return widgets inside those panels, so I'll implement it.

Groogy:
Thank you so much!

> I seem to have missed the notification about this, I'm only seeing this now.

No worries this was still faster response than I expected :)

PR #165, opened 2022-01-01
Update get to also search widget in SubwidgetContainer

Kvaz1r:
Fix for: #164

### DIFF
--- a/src/Container.cpp
+++ b/src/Container.cpp
@@ -313,6 +313,16 @@ namespace tgui
                 if (widget != nullptr)
                     return widget;
             }
+            else
+            {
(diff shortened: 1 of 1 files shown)


Kvaz1r:
Funny, was a bit late :D
Closed in favor - https://github.com/texus/TGUI/commit/3babeca407345ae601aa813fa65f94b0d3ddaafc

texus:
A few minutes too late, but still appreciated. It's good to know that the issue would have also been fixed even if I hadn't accidentally looked at github and thought "why does it say 3 open issues instead of 2?"

PR #171, opened 2022-06-29
Update DataIO.cpp

Gisteron:
As it stood, the attribute inheritance feature did not work for arbitrary sections. Even the theme Black.txt used in an example could not be parsed, and the program would exit with an exception once the first colon character was encountered. I suspect this fix to the parseSection function should allow it to recognize the inheritance operator in theme files.

### DIFF
--- a/src/Loading/DataIO.cpp
+++ b/src/Loading/DataIO.cpp
@@ -408,15 +408,15 @@ namespace tgui
                 }
 
                 REMOVE_WHITESPACE_AND_COMMENTS(true)
-                if (stream.peek() == '{')
+                if (stream.peek() == '=')
                 {
-                    const String error = parseSection(stream, sectionNode, word);
+                    const String error = parseKeyValue(stream, sectionNode, word);
                     if (!error.empty())
                         return error;
                 }
-                else if (stream.peek() == '=')
+                else if (stream.peek() == '{')
                 {
-                    const String error = parseKeyValue(stream, sectionNode, word);
+                    const String error = parseSection(stream, sectionNode, word);
                     if (!error.empty())
                         return error;
                 }
@@ -503,10 +503,10 @@ namespace tgui
             }
 
             REMOVE_WHITESPACE_AND_COMMENTS(true)
-            if (stream.peek() == '{')
-                return parseSection(stream, root, word);
-            else if (stream.peek() == '=')
+            if (stream.peek() == '=')
                 return parseKeyValue(stream, root, word);
+            else if (stream.peek() == '{')
+                return parseSection(stream, root, word);
             else if (stream.peek() == ':')
                 return parseInheritance(stream, root, word);
             else
--- a/src/Loading/DataIO.cpp
+++ b/src/Loading/DataIO.cpp
@@ -420,6 +420,12 @@ namespace tgui
                     if (!error.empty())
                         return error;
                 }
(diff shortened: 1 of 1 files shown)


texus:
Thanks for the PR. I'm a bit confused about what it fixes exactly though.

I don't get any errors when loading the Black theme, is if failing to load a specific widget?
As far as I can tell themes don't have subsections, so this change shouldn't affect loading theme files.

What this change does affect is form files (which also use DataIO). It would allow inheriting a renderer of a widget from a global renderer section in that form file. So I'm still willing to implement this change for that reason (after swapping the place of parseSection and parseInheritance in the file to make the code compile).

Gisteron:
Hmm, well, perhaps the error was someplace else, then. I figured I would first try and compile the many widgets example with SFML backend. One error was that the `tgui` namespace contains no `Canvas` class that the example references, but that was an easy manual fix, be it with replacing all instances of it with `CanvasSFML` or with a local using directive or typedef. After that addition the example would compile, all right, but once the theme is initiated with `tgui::Theme theme{"Black.txt"};`, that's when the program would exit with the message `TGUI Exception: Error while parsing input at line 10. Expected '{' or '=', found ':' instead.`. So I went to the source to look for any place that prints out this error. Perhaps something about the theme constructors/loader changed since that example's publication and this is not a robust way of initiating themes anymore? Maybe I should have opened an issue rather than a PR at this stage. :sweat_smile:

texus:
> One error was that the tgui namespace contains no Canvas class that the example references

The example code on the website still used the Canvas class by mistake (the canvas was removed from the "examples" folder in the TGUI download some time ago but the website wasn't updated). I have now updated the example on the website to solve the compile error with the Canvas class.

> Expected '{' or '=', found ':' instead

I've downloaded and compiled an unmodified TGUI 0.10 version to make sure that I'm testing the correct code. I'm not able to reproduce the exception when loading Black.txt, it loads fine. Is there any chance that you are still linking to older TGUI libraries (e.g. 0.10-beta) while using a newer theme file (since inheritance was only recently added)?

> What this change does affect is form files (which also use DataIO)

The changes proposed in this PR (with some improvements) have just been added in 67c02d8f90d803434672b3dc0b1c27f77064f185, but this change is unrelated to the problem you have.

Gisteron:
> I've downloaded and compiled an unmodified TGUI 0.10 version to make sure that I'm testing the correct code. I'm not able to reproduce the exception when loading Black.txt, it loads fine. Is there any chance that you are still linking to older TGUI libraries (e.g. 0.10-beta) while using a newer theme file (since inheritance was only recently added)?

What my system says I installed is `libtgui-0.10-dev/jammy,now 0.10.0+jammy+dev20220319 amd64` from the `ppa:texus/tgui` repository, which, now that I review the downloads section on the homepage, is apparently indeed a beta variant, and I suppose a somewhat out of date one, at this point. There is no second version installed and I compile with g++ 11.2.0 and link to tgui with `-ltgui` (and sfml linkings as per usual).

Alright, so looks like either I'm looking for beta-compliant theme examples I'll find under releases, or an update to the PPA, if one is in the planning sometime soon. Thanks for your assistance. :)

texus:
The PPA is indeed the beta version.
If you downloaded that version then you can find the compatible Black.txt file in /usr/share/tgui-0.10/themes/

PR #176, opened 2022-12-08
Adding undo functionality to Gui Builder

racebennnett:
Added "Ctrl + Z" Undo command for deletions and moves. More functionality can be added later.

### DIFF
--- a/gui-builder/include/Form.hpp
+++ b/gui-builder/include/Form.hpp
@@ -63,6 +63,8 @@ public:
     bool hasFocus() const;
     void load();
     void save();
+    tgui::String saveState();
+    bool loadState(tgui::String);
     void updateAlignmentLines();
 
 private:
@@ -73,6 +75,7 @@ private:
     void onDrag(tgui::Vector2i mousePos);
     void selectWidget(std::shared_ptr<WidgetInfo> widget);
     std::vector<std::pair<tgui::Vector2f, tgui::Vector2f>> getAlignmentLines() const;
+    bool onDragSaved;
 
 private:
 
--- a/gui-builder/include/GuiBuilder.hpp
+++ b/gui-builder/include/GuiBuilder.hpp
@@ -42,14 +42,14 @@ public:
 
     enum class eUndoType
     {
-        Delete, Move
+        Delete, Move, Paste, SendtoFront, SendtoBack, CreateNew
     };
 
     void reloadProperties();
     void widgetSelected(tgui::Widget::Ptr widget);
     void formSaved(const tgui::String& filename);
     void closeForm(Form* form);
-    void undoWidgetSave(eUndoType);
+    void saveUndoState(eUndoType);
 
 private:
 
@@ -89,7 +89,7 @@ private:
     void pasteWidgetRecursive(const CopiedWidget& copiedWidget, tgui::Container* parent);
     void copyWidgetToInternalClipboard(std::shared_ptr<WidgetInfo> widgetInfo);
     void pasteWidgetFromInternalClipboard();
-    void undoWidgetLoad();
+    void loadUndoState();
 
     void widgetHierarchyChanged();
     void updateSelectedWidgetHierarchy();
@@ -156,6 +156,7 @@ private:
 
     std::vector<tgui::String> m_undoSaves;
     std::vector<tgui::String> m_undoSavesDesc;
+    unsigned int m_undoSaveMaxSaves;
 
 };
 
--- a/gui-builder/include/GuiBuilder.hpp
+++ b/gui-builder/include/GuiBuilder.hpp
@@ -64,6 +64,7 @@ private:
         std::vector<CopiedWidget> childWidgets;
     };
 
+
     bool loadGuiBuilderState();
     void saveGuiBuilderState();
     void loadStartScreen();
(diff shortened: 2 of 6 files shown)


texus:
Thanks for adding this.

I haven't tested the code yet, but after looking at the code I have two remarks:
- saveState probably shouldn't call setChanged because it would think the form has been saved to disk and won't warn when closing the gui builder without saving the form
- I would rename undoWidgetSave/undoWidgetLoad, it sounds like the function is undoing a save or load that happened before. Maybe something like saveWidgetUndoState (or without the State at the end), or just saveUndoState (since it's the state of the form and not really a widget that you are saving)?

There are some other minor style issues in the changes, but I will clean those up myself when merging the PR. Do you intent to still add more in the near future? If so then I will just keep this PR open as WIP for some time, that would reduce the work in merging back and forth changes.

The "android-sdl" CI test is failing, but you can ignore that as it unrelated to your changes. Due to an OS update and a mistake in my script it is trying to use c++17 on a compiler that only supports c++14

racebennnett:
I will remove setChanged from saveState as you mentioned as it doesn't need to be called, although I tested it and I still got close warnings after it had been run. 

I will rename undoWidgetSave/undoWidgetLoad to saveUndoState/loadUndoState. I agree the widget shouldn't be in the name.

Leave it open for now. It works functionally for me so I wasn't going to make anymore changes until I knew it was going to get added in and it was worth my time.

Thanks

racebennnett:
Well I think I got it about where I want it. The undo nearly covers everything, and the load feature no longer leaks memory of any significance. However, now there is a build error with some builds due to using tryemplace (C++17). Do I need to fix this or can the builds be updated?

texus:
TGUI still needs to compile with c++14, so c++17 code can't be used (unless it is placed behind and ifdef and an alternative c++14 implementation is provided). So the CI builds won't be changed.

I'm planning to review (and probably merge) your code tomorrow. If I still find issues then I will fix them during the merge.

texus:
Thanks again for contributing this functionality.

I improved a few things while reviewing the code:
- Resizing the widget by dragging the sides or corners wasn't supported by undo
- Resizing widgets by pressing shift + arrow key was identified with undo type "Move"
- Pressing Ctrl+Z incorrectly caused the form to be marked as saved (because a completely new form is created by loadUndoState)
- The widget now remains selected while undoing.
- Various code style changes and code simplifications.

For future reference, here is a list of remaining issues that I'm aware of:
- Moving or resizing widgets with arrow keys causes many states to be created, undo reverts pixel by pixel. After the first state is stored, the gui builder should probably keep track of whether the control and shift keys are pressed and not store any more states until they are released (similar to how dragging only stores the state once when dragging begins).
- I don't like how loadForm is called from loadUndoState. The majority of the code executed by loadUndoState is often not even needed. Fixing this seems to be non-trivial though.
- Making a change and undoing it now leaves the form marked as changed/unsaved. Fixing this requires keeping track of where in the list of undo states the last form save occurred, but I don't believe it is important enough to put effort into this.

racebennnett:
Thanks for the improvements and reforms.
- For the moving pixel by pixel I am impartial. If you overstep one pixel you can always undo to get back to where you wanted, but I understand what you are saying.
- I debated whether to use loadForm. The reason I did end up using it was to ensure any future changes that needed loading would be used for both "load" types to prevent having to duplicate everything, but yes it certainly could be split.

I would also like to add that with your update you have added the memory leaks back into program with:
 valueComboBox->onItemSelect([onChange,comboBox=valueComboBox]{ onChange(comboBox->getSelectedItem()); });
 valueComboBox->onItemSelect([onChange,comboBox=valueComboBox]{ onChange(comboBox->getSelectedItem()); });
valueEditBox->onReturnOrUnfocus([onChange,editBox=valueEditBox.get()] { onChange(editBox->getText()); });
.

These 3 lines are just getting another shared pointer to themselves and will never delete themselves.

texus:
While porting the changes to 0.10 I found and (partially) fixed two more issues:
- Ctrl+Z didn't work when the properties window was focused (which is the case if you just changed one of the properties).
- An undo state was created when the form was resized, but the size wasn't actually stored because saveWidgetsToStream only saves the child widgets. Any time ctrl+z was pressed the form restored to the size it had when it was last saved (even when undoing something other than the form size). I've patched this so that the form size is no longer reset during an undo action, but you can also no longer undo a form size change at the moment.

> These 3 lines are just getting another shared pointer to themselves and will never delete themselves.

The 3th line should be fine (based on my understanding of the issue, I didn't test it), but I did indeed forget to add `.get()` for the others.

racebennnett:
You are correct, the ".get()" was indeed the issue. I verified it no longer leaks with that method. 

Honestly I'm not entirely sure why the get() solves the issue. I thought it was still a shared pointer reference.

texus:
The `get()` function works because it returns a raw pointer to the resource. So the lambda will store a `tgui::EditBox*` instead of a `tgui::EditBox::Ptr`. This could be dangerous as well because it would crash if the lambda function is called after the widget is destroyed, but that will never happen in these cases as it is impossible to get a callback from a widget that doesn't exist anymore.

In general a safer method might be `valueEditBox->onReturnOrUnfocus([editBox=std::weak_ptr<tgui::EditBox>(valueEditBox)] { /* use editBox->lock() here */ });`, but I've never been a big fan of using weak_ptr (because it takes more code than a raw pointer and I usually don't get in situations where a pointer could outlive the shared_ptr resource).

In the 0.10 branch I've removed the majority of `[=]` captures and decided to pass all widgets with the `.get()` method (https://github.com/texus/TGUI/commit/101a5fa2972489a7031804335c4181573234234a). This should reduce the chance of other memory leaks, and maybe even fix some small existing ones.

racebennnett:
Thanks for the explanation. I was unaware of the raw pointer retrieval with get(). I agree with you here that the risk is non existent  due to the parent nature of the object that is being referenced.

PR #195, opened 2023-05-16
Add missing include in include/TGUI/Utf.hpp

TedLyngmo:
When building with `g++ 13.1.1` there is a compilation error because the definition of `std::uint8_t` is missing.

This patch adds `#include <cstdint>`.


vfjpl:
Can this be backported to 0.8 branch? :)

PR #203, opened 2023-08-26
Fixed widgets not being updated in BoxLayout when a child widget's index was changed using setWidgetIndex (fixes #202)

### LINKED ISSUE #202: Setting a widget's index in a layout doesn't seem to update it until the layout is resized
I've come across an issue with updating widget indices whilst developing my game, and I have been able to replicate it in isolation with the following code:

```cpp
#include <TGUI/TGUI.hpp>
#include <TGUI/Backend/SFML-Graphics.hpp>
#include <iostream>

using namespace tgui;

int main() {
    sf::RenderWindow window{ {800, 600}, "TGUI example - SFML_GRAPHICS backend" };
    Gui gui{ window };

    auto layout = VerticalLayout::create();
    gui.add(layout);

    auto button1 = Button::create("Button1");
    button1->setPosition("50px", "50px");
    button1->setSize("50px", "50px");
    layout->add(button1);

    auto button2 = Button::create("Button2");
    button2->setPosition("75px", "75px");
    button2->setSize("50px", "50px");
    layout->add(button2);

    auto button3 = Button::create("Button3");
    button3->setPosition("100px", "100px");
    button3->setSize("50px", "50px");
    layout->add(button3);

    layout->setWidgetIndex(button3, 0);

    // Adding a new widget to the layout updates it.
    //auto button4 = Button::create("Button4");
    //layout->add(button4);

    // As does resizing the layout.
    //layout->setSize("50%", "50%");

    while (window.isOpen()) {
        sf::Event event;
        while (window.pollEvent(event)) {
            gui.handleEvent(event);

                window.close();
        }
        window.clear();
        gui.draw();
        window.display();
    }
}
```

The expected outcome is that the buttons will be displayed in the following order, from top to bottom:

1. Button3
2. Button1
3. Button2

However, the buttons are instead displayed in their original order:

1. Button1
2. Button2
3. Button3

*That is*, until the layout is resized, either by manually resizing the window, resizing the layout itself programmatically (e.g. `setSize()`), or adding a widget to the layout to invoke a resize.

I can also replicate this issue with a `HorizontalLayout` and `HorizontalWrap`, but I cannot replicate it with any other container type. You can see that if you change `VerticalLayout` to `Group`, for example, Button2 will appear on top of the other two buttons, which is the intended outcome.

Can replicate using the latest changes on the `1.x` branch (currently commit `6c4879e`). Seems like this could be a simple fix so I'll do some digging now and report any findings and/or make a PR if I make any progress.

CasualYT31:
Previously, when a widget within a `BoxLayout`-derived container (`HorizontalLayout`, `VerticalLayout`, `HorizontalWrap`) had its index changed via `Container::setWidgetIndex()`, it would not update the widgets in the `BoxLayout` using `updateWidgets()`. This would cause the widget to appear not to have been assigned a new index, until `updateWidgets()` was called in some other way (for example, when the `BoxLayout` gained a new widget, or if the `BoxLayout` was resized).

Now, `Container::setWidgetIndex()` is virtual, and `BoxLayout` overrides it. The overridden implementation invokes `Container::setWidgetIndex()`, and returns the result of the call. If the result of that call is `true`, the index change was successful, so `updateWidgets()` is called to show the change in the index immediately.

### DIFF
--- a/include/TGUI/Container.hpp
+++ b/include/TGUI/Container.hpp
@@ -345,7 +345,7 @@ TGUI_MODULE_EXPORT namespace tgui
         ///
         /// @return True when the index was changed, false if widget wasn't found in the container or index was too high
         /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
-        bool setWidgetIndex(const Widget::Ptr& widget, std::size_t index);
+        virtual bool setWidgetIndex(const Widget::Ptr& widget, std::size_t index);
 
 
         /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/include/TGUI/Widgets/BoxLayout.hpp
+++ b/include/TGUI/Widgets/BoxLayout.hpp
@@ -138,6 +138,20 @@ TGUI_MODULE_EXPORT namespace tgui
 
 
         /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+        /// @brief Changes the index of a widget in this container
+        ///
+        /// Widgets are drawn in the order of the list, so widgets with a higher index will appear after others.
+        ///
+        /// @param widget  Widget that is to be moved to a different index
+        /// @param index   New index of the widget, corresponding to the widget position after the widget has been moved
+        ///
+        /// @return True when the index was changed, false if widget wasn't found in the container or index was too high
+        /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+        virtual bool setWidgetIndex(const Widget::Ptr& widget, std::size_t index) override;
+        using Container::setWidgetIndex;
+
+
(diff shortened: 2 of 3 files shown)


texus:
Thanks for the detailed information and the PR with a fix.

One minor detail: the `using Container::setWidgetIndex;` line isn't needed. I assume you added it because it was used on other overridden functions in BoxLayout, but it is only needed in a special case. If the base class contains multiple functions with the same name (but with different parameters), and only one of them is overridden, then the others would become hidden. The `using` line makes it such that you can still call both the overridden (e.g. `setSize(Layout2d)`) and non-overridden (e.g. `setSize(Layout, Layout)`) functions. There is only one `setWidgetIndex` function, so it doesn't apply here and the `using` line has no effect.

PR #204, opened 2023-08-29
Add getSignal() to TabContainer

CasualYT31:
The `TabContainer` did not have an implementation for `Widget::getSignal()`. This caused invoking `getSignal("SelectionChanged")` and `getSignal("SelectionChanging")` on `Widget` objects to throw exceptions.

Now, both invocations can be made and signal handlers can be connected. `Widget` signals can still be accessed, too.


texus:
Thanks. This issue also seems to exist in SpinControl, could you perhaps add another commit to fix it there as well?

CasualYT31:
> Thanks. This issue also seems to exist in SpinControl, could you perhaps add another commit to fix it there as well?

Sure, I'll get it done within the next few hours. Should've checked to be fair.

PR #205, opened 2023-08-30
Proposal: onCaretPositionChanged signal added to EditBox and TextArea

CasualYT31:
This can be useful for implementing a status bar in a text editing program that displays the caret position, for example.

I've added the signal to both `EditBox` and `TextArea`. `EditBox::onCaretPositionChanged` accepts an optional `size_t` parameter storing the new position. `TextArea::onCaretPositionChanged` accepts an optional `Vector2<size_t>` parameter that stores the new line and column number of the caret.

I realise that the `m_selEnd` `Vector` in `TextArea` is not currently publicly accessible: would this mean that the caret position should be emitted as a `size_t` for `TextArea`, or should some method/s be added to access this information (e.g. `getCaretLineAndColumn()`)?

I've done some basic testing and it works quite well so far. Some actions (such as pasting from the clipboard whilst text is selected) will emit multiple signals, but otherwise, a signal will only be emitted once, when the caret position actually changes. Would it be worth preventing the emission of the new signal in such cases or can these be ignored?

Let me know if this addition wouldn't be suitable, or if I'll need to change my approach.


texus:
> I realise that the m_selEnd Vector in TextArea is not currently publicly accessible: would this mean that the caret position should be emitted as a size_t for TextArea, or should some method/s be added to access this information (e.g. getCaretLineAndColumn())?

Right now the interface only exposes the `size_t` caret, because the word-wrapping result isn't public. You have a text in your code that you pass to the setText function or receive via the getText function, and the size_t caret position provides an offset within that text.

There is no real use for a `Vector2<size_t>` caret as you don't know which text is positioned on which line, so you can't use the caret for anything except just printing it's value (which is indeed what you might find in a status bar).

While it makes more sense to me to emit `size_t` instead of `Vector2<size_t>` in the current API, calculating the size_t caret each time onCaretPositionChanged is called would be wasteful (especially since most people won't have a callback connected). So for now I would just make onCaretPositionChanged a `Signal` which does not have a parameter.

You can add `getCaretLineAndColumn()` so that you still have a way to get the line and column to print in the status bar, and you can just call that function from inside the onCaretPositionChanged callback instead of receiving it via a parameter.

> Some actions (such as pasting from the clipboard whilst text is selected) will emit multiple signals.
> Would it be worth preventing the emission of the new signal in such cases or can these be ignored?

Those can be ignored, I don't see multiple signals as a problem. However I do care that the caret position is valid each time the signal is triggered. If inside the onCaretPositionChanged callback you access getText() then the caret position must be always be smaller or equal to the text length. So on places where both the text and the caret position are changed (such as deleting selected text), the text should always be updated before the onCaretPositionChanged is triggered.

This will probably already be the case on most places, and maybe even everywhere, but it is worth checking. There has to be no relation between the onCaretPositionChanged and onTextChanged signals though, that would be too complicated. It's only important that m_text is updated before calling onCaretPositionChanged. It doesn't matter whether onCaretPositionChanged is called before or after onTextChanged and it doesn't matter that onCaretPositionChanged will be called more often than onTextChanged.

CasualYT31:
Going to keep things simple by declining this PR and creating two separate PRs for both `EditBox` and `TextArea`. `TextArea`'s proving to be more complicated than I anticipated. Thanks for your guidance and I'll take it on board.

PR #207, opened 2023-08-31
TextArea: Add getCaretLine(), getCaretColumn(), and onCaretPositionChange Signal. String: Add count() methods

CasualYT31:
# Changes to Public Interfaces

- Add `count(char, size_t)` methods to `String`. These methods count the number of occurrences of a given character within the string.
- Add `getCaretLine()` and `getCaretColumn()` to `TextArea`.
- Add `onCaretPositionChange` signal to `TextArea`.

# Testing

Most of the tests I performed can be seen below. In all cases, `getText()` returned the correct/updated text when invoked within the `onCaretPositionChange` signal handler. "Location" or "position" refers to the current caret position unless otherwise stated.

| Test Case | Actions | Result |
| --- | --- | --- |
| `setSelectedText()` | Select portion of the text. | Emitted once. |
| `setSelectedText()` | Select portion of the text that ends at the current caret position. | Emitted once, despite caret not changing position. This case probably doesn't matter. |
| `setSelectedText()` | Select the same portion of the text. | Not emitted. |
| `setCaretPosition()` | Set to a different location. | Emitted once. |
| `setCaretPosition()` | Set to the same location. | Not emitted. |
| `leftMousePressed()` | Single click at different location. | Emitted once. |
| `leftMousePressed()` | Single click at same location. | Not emitted. |
| `leftMousePressed()` | Clicking on scrollbars. | Not emitted. |
| `leftMousePressed()` | Double click. | Emitted once for each click, unless the caret position does not change on a click. |
| `mouseMoved()` | No mouse button held. | Not emitted. |
| `mouseMoved()` | Mouse button held. | Emitted once for every change of the caret position. |
| `keyPressed()` | Left, Right, Up, Down, End, Home, End + Ctrl, Home + Ctrl. | Emitted once per key press, but only when caret position changes. |
| `keyPressed()` (`moveCaretWordBegin()`, `moveCaretWordEnd()`, `moveCaretPageUp()`, `moveCaretPageDown()`) | Page Up, Page Down, Left + Ctrl, Right + Ctrl. | Emitted once per key press, but only when caret position changes. |
| `textEntered()` | Vertical scrollbar is Automatic or Always. Enter key. | Emits one signal per key press. |
| `textEntered()` | Vertical scrollbar is Automatic or Always. Any other key. | Emits one signal per key press. |
| `textEntered()` | Vertical scrollbar is Never. Enter key, and it can fit. | Emits one signal per key press. |
| `textEntered()` | Vertical scrollbar is Never. Enter key, and it can't fit. | Not emitted. |
| `textEntered()` | Vertical scrollbar is Never. Any other key, and it can fit. | Emits one signal per key press. |
| `textEntered()` | Vertical scrollbar is Never. Any other key, and it can't fit. | Not emitted. |
| `deleteSelectedCharacters()` | End Sel > Start Sel | Since End Sel is reset to Start Sel, emits one signal. |
| `deleteSelectedCharacters()` | End Sel <= Start Sel | Since End Sel is not changed, no signal is emitted. |
| `backspaceKeyPressed()` | At start of text. | Not emitted. |
| `backspaceKeyPressed()` | At any other location, no text selected. | Emitted once. |
| `backspaceKeyPressed()` | At beginning of line that was added due to word wrap. | Emitted once. |
| `pasteTextFromClipboard()` | Pasting empty string. | Not emitted. |
| `pasteTextFromClipboard()` | Pasting string with no text selected, regardless of location. | Emitted once. |
| `pasteTextFromClipboard()` | Pasting string with text selected, caret at start of selection. | Emitted once. |
| `pasteTextFromClipboard()` | Pasting string with text selected, caret at end of selection. | Emitted twice, once after deleting the selected characters, once after pasting string. |
| `selectAllText()` | Positioned at end of text. | Not emitted. |
| `selectAllText()` | Positioned anywhere else. | Emitted once. |

### DIFF
--- a/include/TGUI/String.hpp
+++ b/include/TGUI/String.hpp
@@ -873,6 +873,11 @@ TGUI_MODULE_EXPORT namespace tgui
         TGUI_NODISCARD inline bool ends_with(char16_t ch) const noexcept;
         TGUI_NODISCARD inline bool ends_with(char32_t ch) const noexcept;
 
+        TGUI_NODISCARD std::size_t count(const char ch, const std::size_t pos = npos) const noexcept;
+        TGUI_NODISCARD std::size_t count(const wchar_t ch, const std::size_t pos = npos) const noexcept;
+        TGUI_NODISCARD std::size_t count(const char16_t ch, const std::size_t pos = npos) const noexcept;
+        TGUI_NODISCARD std::size_t count(const char32_t ch, const std::size_t pos = npos) const noexcept;
+
         inline friend bool operator==(const String& left, StringView right);
         inline friend bool operator==(const String& left, const char32_t* right);
         inline friend bool operator==(const String& left, const std::u32string& right);
--- a/include/TGUI/String.hpp
+++ b/include/TGUI/String.hpp
@@ -873,10 +873,10 @@ TGUI_MODULE_EXPORT namespace tgui
         TGUI_NODISCARD inline bool ends_with(char16_t ch) const noexcept;
         TGUI_NODISCARD inline bool ends_with(char32_t ch) const noexcept;
 
-        TGUI_NODISCARD std::size_t count(const char ch, const std::size_t pos = npos) const noexcept;
-        TGUI_NODISCARD std::size_t count(const wchar_t ch, const std::size_t pos = npos) const noexcept;
-        TGUI_NODISCARD std::size_t count(const char16_t ch, const std::size_t pos = npos) const noexcept;
-        TGUI_NODISCARD std::size_t count(const char32_t ch, const std::size_t pos = npos) const noexcept;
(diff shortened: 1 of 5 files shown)


CasualYT31:
Oh, I should probably at least add tests for the new `String::count()` method. I've figured out how to get tests working so I'll work on those next.

CasualYT31:
In doing so, I've realised that `pos` in `count()` doesn't really match with how such a parameter works in the rest of the class, e.g. `find()`. Since both methods traverse forwards, it is supposed to represent the first character to "find" or "count" from, and not the final character to "find" or "count" to.

Even considering this, I've ended up introducing another overflow bug in `count()`. So I'll need to rework this PR a tiny bit.


Have pushed the fixes now :+1:

PR #208, opened 2023-08-31
SubwidgetContainer Inherit Font

CasualYT31:
Previously, a `TabContainer` would not pass on its font to its internal `Container`. This would mean that it would be stuck using the default font unless it was set manually.

Now, `TabContainer` overrides `rendererChanged()`, and sets its `Container`'s inherited font to the `TabContainer`'s font once it is notified that its font has changed.

I noticed the same problem was also present for the `SpinControl`, so now its font is passed on to its `EditBox`, too.

I am not too sure if more properties should be handled in these methods. If you can think of any more let me know and I can add them.

### DIFF
--- a/src/SubwidgetContainer.cpp
+++ b/src/SubwidgetContainer.cpp
@@ -225,6 +225,12 @@ namespace tgui
         }
 
         Widget::rendererChanged(property);
+
+        // Make sure to pass on the SubwidgetContainer's font to the internal container
+        if (property == U"Font")
+        {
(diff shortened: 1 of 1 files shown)


texus:
Could this perhaps be solved in `SubWidgetContainer`, i.e. by overriding rendererChanged there and passing the font to `m_container` so that no code needs to be changed in `TabContainer` and `SpinControl`?

CasualYT31:
Good point, got a bit ahead of myself. I'll clean the code up.

PR #211, opened 2023-09-02
TextArea: Provide option to customise what is inserted when Tab is pressed

CasualYT31:
# Overview

* Refactored string insertion code.
* Pasting now abides by maximum characters limit (as does pressing tab).
* Updated tests for both Tab Entered and Copy and Paste.
* Added tests for new `setTabString()` and `getTabString()` methods.
* Remove unimplemented `TextArea::selectText()` method.
* Prevent emitting two identical `onCaretPositionChange` signals when pasting/inserting text in certain cases (namely if the pasted string includes a newline).

# Pasting Text

Previously, pasting text into a `TextArea` could cause the text to exceed the maximum character limit. I'm assuming this is unintended behaviour, so I've now introduced code that will truncate the pasted text if inserting it would cause the text to exceed the limits. I do this by removing characters from the end of the pasted string. LMK if this isn't desired, or if a different truncation approach should be taken.

# Customise Tab

This can be handy when the user wants to insert spaces instead of tabs, for example. I've left the default to `\t`, but the user can set any size string now, including an empty string. I've refactored out the pasting code into a separate method so that I could reuse it for Tab key events, so it will follow the same maximum characters limit logic, too. `TextArea` makes sure to remove cartridge return (`\r`) characters from the string given to `setTabString()` before assigning it internally, and I've included a note in the docstring warning the user of this.


include/TGUI/Widgets/TextArea.hpp
```diff
@@ -219,6 +219,30 @@ TGUI_MODULE_EXPORT namespace tgui
         TGUI_NODISCARD std::size_t getMaximumCharacters() const;
 
 
+        /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+        /// @brief Changes the string that is inserted when the Tab key is pressed
+        ///
+        /// @param tabtext  The string to insert when Tab is pressed.
```
texus: `tabtext` should be `tabText`

include/TGUI/Widgets/TextArea.hpp
```diff
@@ -219,6 +219,30 @@ TGUI_MODULE_EXPORT namespace tgui
...
+        ///
+        /// Defaults to "\t".
+        ///
+        /// @warning Any cartridge return characters given in this string (\r) will be removed!
+        ///
```
texus, line 230: Perhaps you can also add a warning here that the tab key doesn't affect the text area unless `gui.setTabKeyUsageEnabled(false)` was called.

texus (commented):
I'm not entirely certain how useful controlling the tab key is, but I'm ok with adding this.
If you can still make the minor fix to the documentation of setTabString then I'll merge this.

PR #212, opened 2023-09-02
Widgets with dots in their names can now be fully referenced in layout expressions

CasualYT31:
Previously, it was impossible to reference a widget's X, Y, width, etc. properties in a layout expression if that widget had dots in its name.

Finding the last dot instead of the first one resolves this issue.

There might well be edge cases (such as if there is a widget called `Widget.x`) but dealing with these is not worth it, and the user shouldn't be naming widgets in such a way if they need to access these properties reliably.

### DIFF
--- a/include/TGUI/Backend/Window/BackendGui.hpp
+++ b/include/TGUI/Backend/Window/BackendGui.hpp
@@ -229,7 +229,10 @@ TGUI_MODULE_EXPORT namespace tgui
         /// @param widgetPtr   Pointer to the widget you would like to add
         /// @param widgetName  If you want to access the widget later then you must do this with this name
         ///
-        /// @warning The widget name should not contain whitespace
+        /// @warning Widgets should be named as if they are C++ variables, i.e. names must not include any whitespace, or most
+        ///          symbols (e.g.: +, -, *, /, ., &), and should not start with a number. If you do not follow these rules,
+        ///          layout expressions may give unexpected results. Alphanumeric characters and underscores are safe to use,
+        ///          and widgets are permitted to have no name.
         ///
         /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
         void add(const Widget::Ptr& widgetPtr, const String& widgetName = "");
--- a/include/TGUI/Container.hpp
+++ b/include/TGUI/Container.hpp
@@ -133,7 +133,10 @@ TGUI_MODULE_EXPORT namespace tgui
         /// @param widgetPtr   Pointer to the widget you would like to add
         /// @param widgetName  You can give the widget a unique name to retrieve it from the container later
         ///
-        /// @warning The widget name should not contain whitespace
+        /// @warning Widgets should be named as if they are C++ variables, i.e. names must not include any whitespace, or most
+        ///          symbols (e.g.: +, -, *, /, ., &), and should not start with a number. If you do not follow these rules,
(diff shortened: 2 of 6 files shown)


CasualYT31:
I saw that the checks were failing and panicked because I forgot to run the tests. So I did and this test failed:

```
            button2->setPosition({"parent.b1.parent.b1.x"}, {"&.b1.&.b1.y"});
            REQUIRE(button2->getPosition() == tgui::Vector2f(40, 60));
```

Probably because my change would cause the `Layout` to try and find a non-existent widget called `parent.b1.parent.b1`. I'll see if adding a check to guard against this case resolves it.

texus:
I'm not sure if this should be supported (if it is even possible). You can have a layout containing "MainPanel.SubPanel.Button.x" and it would be impossible to know whether the widget you are looking for is "MainPanel.SubPanel" that contains a "Button" or whether the widget you need to look for is "MainPanel". You might be able to check which widgets exist, but the layout may be set prior to the existence of the bound widget (at least that used to be the case, I'm not certain if that is still allowed as I did have some prior issues with that).

Is there any reason why you must use dots and can't e.g. use underscores?

EDIT: I checked the documentation and it only mentions that the name shouldn't contain whitespace (and that note only exists in Container::add and not at Widget::setWidgetName). The idea was to only allow simple names though, so with the existing code the documentation should have stated that the name shouldn't contain whitespace, brackets, quotes or dots.

CasualYT31:
Good points, if I had known about this when I started naming my widgets in this way I definitely would've used underscores. I don't have any reason to avoid using underscores, so in the near future, I'll work on updating that in my own code. I've never had to rely on complex expressions yet so I never ran into issues with using `rfind`.

If one learns about layout expressions before using the library a lot, then they'd know not to name their widgets in this way. But I think I only became aware of them after I had already implemented a widget naming system using dots, and by then it was just easier to make this change to get the expressions I used to work. I thought that this was just a "bug" at first but I was wrong about that.

I guess it would be helpful to include warnings next to `setWidgetName()` methods saying that dots shouldn't be included in them if you rely on layout expressions to work correctly. I can update this PR to add these warnings if you think it would be a good idea?

texus:
> I can update this PR to add these warnings if you think it would be a good idea?

That would be great. The warning needs to be added to `Widget::setWidgetName` and the same warning can be put at `Container::add` (where it replaces the existing warning). Now that I think of it, there are many things that probably shouldn't be used in the name. A `-` for example could be wrongly considered as a minus instead of it being part of the name. So the name should basically be limited to similar rules as c++ variable names, but I'm not sure how to best describe that.

CasualYT31:
Alright :+1: I've decided upon this warning:

```cpp
        /// @warning Widgets should be named as if they are C++ variables, i.e. names must not include any whitespace, or most
        ///          symbols (e.g.: +, -, *, /, ., &), and must not start with a number. If you do not follow these rules, layout
        ///          expressions may give unexpected results. Alphanumeric characters and underscores are safe to use, and
        ///          widgets are permitted to have no name.
```

(I've not actually tested if you can start a widget name with a number, but I'm assuming you can't).

LMK if this should be worded better.

texus:
Technically there isn't a reason why the name can't start with a number, but it's probably still better to avoid doing that. Maybe change `and must not start with a number` to `and should not start with a number` to put less emphasis on this part?
Other than that it's better than what I was going to write.

PR #216, opened 2023-09-02
TextArea: improvements to setSelectedText()

CasualYT31:
* A field called `m_retainSelEnd` was introduced in 3f7fb75a which prevented `setSelectedText()` from changing `m_selEnd` more than once using `setCaretPosition()`. This has been replaced with a separate `setCaretPositionImpl()` method with `bool` parameters.

I did this to avoid the really small chance of the second call to `setCaretPosition()` throwing for whatever reason, leaving `m_retainSelEnd` set to `true`, meaning the user would no longer be able to change the `m_selEnd` using `setCaretPosition()`. Using a bool flag in this way is kinda smelly IMO so I removed it.

* `setSelectedText()` will now only emit `onCaretPositionChange` if it actually moved the caret position in all cases.

Previously, a "duplicate" `onCaretPositionChange` signal would be emitted if `setSelectedText()` selected a new portion of text that ended at the current caret location/end selection. This no longer happens.


texus:
This does look cleaner compared to the bool flag.

Now that you turned the variable into a bool parameter, it stands out to me that maybe the logic should be inverted, i.e. have a `updateSelEnd` parameter instead of `retainSelEnd`.

You added helpful comments behind each line to tell what the bool parameters mean, but it feels like the bool values don't match the description. For example for `Update m_selEnd and emit onCaretPositionChange` I would expect 2 `true` values to match that description and for `Replace m_selEnd but do not emit onCaretPositionChange yet` I would expect one value to be `true` and another to be `false`.

PR #240, opened 2024-05-08
Add PanelListBox to gui builder and add missing setItemsWidth function to PanelListBox

Schweini07:
I've noticed that there is no PanelListBox widget in the gui builder, so this PR adds it :slightly_smiling_face:.

I also noticed that PanelListBox has a `getItemsWidth` function, but no `setItemsWidth` function. A `getItemsHeight` and a `setItemsHeight` function exist, so I assume this was just forgotten by the initial PR implementing it. Therefore I added this in a separate commit.


texus:
ItemsWidth was intentionally not implemented. The panels are stacked below each other, and you are free to choose the panel height / item height, but the width is always equal to the list box width.

Schweini07:
Ah, I understand. I've rebased the commit now, so only the gui builder functionality remains :+1:

texus:
I didn't implement the widget in the gui builder myself because there wasn't much you could do at design time, but at least now you can give it a position and size in the gui builder.

The CI failed because I pushed some bad commits around the time you made this PR, but once merged everything should be fine.

PR #241, opened 2024-05-19
Window focus signals

lpogic:
The window focus state change signals added. It is imitating the Widget API (`onFocus`/`onUnfocus`).

There was also a problem with gui refreshing when it was changed in `onUnfocus` callback. Marking the event as processed resolved the issue.


texus:
I'm not sure how I feel about this PR. My idea has always been that you should get window events from the window, the mainLoop() function was only added to reduce the size of example code. Over time it just got more complex because people actually started relying on it working properly for more complex cases.

I don't feel like handleEvent needs to return true for focus events, because the original idea was that true is returned for things that are handled by TGUI widgets and could be ignored by the user.

So I'm going to close this PR and make a commit that implements this slightly differently. I'll add an `onWindowFocus` and `onWindowUnfocus` signal, so that you can continue to use mainLoop() even if you need an unfocus signal. And I'll fix the gui refreshing in the mainLoop function similar to how MouseLeft is handled.

lpogic:
What prompted me to prepare this PR was the need to handle window focus/unfocus in callback (I use TGUI through bindings in my project). I wasn't sure if BackendGui was the right place for these signals, but I couldn't think of anything better.

PR #246, opened 2024-09-26
Add missing include in include/TGUI/Utf.hpp

vfjpl:
https://github.com/texus/TGUI/pull/195


texus:
Is there a good reason why you are still using TGUI 0.8? I'm guessing you are trying to compile some old code?
While 0.8 hasn't officially been deprecated, it hasn't received any more updates in the last 3.5 years. I'm fine with this being merged though.

PR #271, opened 2025-12-24
Small fixes

jjuhl:
Hi Texus

Here's another batch of small fixes/cleanups.
Since I'm using your library in my own project I can't help going through the code to see what I'm using and I then feel compelled to fix up stuff I notice along the way. This is all minor stuff, but IMHO nice-to-have's, but I'll let you be the final judge of that - pull if you like it :-).    If you want changes, just let me know.

Kind regards
Jesper Juhl


texus:
About the "Don't use deprecated headers" commit: does c++ guarantee somewhere that `<cstring>` will include `<string.h>` internally? All CI tests completed, so I guess in practice it is always the case.
Those `.h` variants were included there because those are the ones being included by nanosvg.h and I'm relying on the include guards to match on both places. If string.h isn't included before including nanosvg.h then the code is not valid anymore because I'm wrapping the nanosvg.h include in a namespace (so it isn't allowed to include new std headers).

The "Use anonymous namespace rather than static" is the only change that I don't necessarily believe is objectively better. With `static` in the declaration you can immediately tell that the function has internal linkage while the namespace might not be as visible if you have a a bunch of internal functions. That being said, I'm not against the change to make things more consistent, especially since static structs don't exist so it's a bit weird to use `static` for functions but anonymous namespaces for structs. So using anonymous namespaces for both does make sense even though I typically don't do it for simple functions.

Could you move the brace from `namespace {` to a new line and change `} // namespace` to `} // anonymous namespace`? Then that commit would be perfect too.

A quick code search for "static void" revealed that there are at least some other static free functions in BackendRenderTarget.cpp, so maybe it would make sense to also put those in anonymous namespaces? They are only part of the tgui namespace because they use TGUI classes.

The other changes all look great, thanks.

jjuhl:
Thank you for the feedback.
I don't have time to do anything about it for the next week or so, but then I'll fix it all up based on your feedback.

jjuhl:
Hi Texus,

Plans change and I found some time to look at this a bit earlier than I had anticipated :)

"About the "Don't use deprecated headers" commit: does c++ guarantee somewhere that <cstring> will include <string.h> internally?"
- I don't believe the standard explicitly guarantees that. As far I know, the standard guarantees "behavioral equivalence" and that if a C-style header (like string.h) exists, then a corresponding "cstring" header will also exist. While the string.h header will place symbols in the global namespace, the cstring one will place it in the std:: namespace and whether or not it also gets added to the global namespace is optional. In practice I've never had any problems replacing the C-style version with the C++style version, but I've moved the commit to the tip of the branch so it's easy for you to merge the others and just drop this one commit if you want.

"Could you move the brace from namespace { to a new line and change } // namespace to } // anonymous namespace? Then that commit would be perfect too." 
- Done.

"A quick code search for "static void" revealed that there are at least some other static free functions in BackendRenderTarget.cpp, so maybe it would make sense to also put those in anonymous namespaces?" 
- I believe that would make sense, but I've not done it as part of this commit series since it would require a bit more work than just moving the code. But, I have other changes coming up in the future, so I'll put it on my todo list and you'll get that in a future PR :)

"The other changes all look great"
- Thank you :)

Kind regards,
Jesper

PR #273, opened 2025-12-26
Cmake changes

jjuhl:
Hi Texus

Here a some small changes to your CMakeLists.txt files I have made locally. Merge if you like :-)

/Jesper Juhl


texus:
The standard and extensions are already set on a per target basis (see https://github.com/texus/TGUI/blob/1.x/cmake/Macros.cmake#L92-L94), so setting them globally isn't needed.
Looking at the comment above those lines, your code would probably also break using TGUI as a c++20 modules in some cases.

jjuhl:
I've dropped that part of the branch.

PR #275, opened 2025-12-29
Warning fixes

jjuhl:
Hi Texus

Another tiny PR with a trivial fix.
This was found by adding an extra GCC warning flag.

Kind regards
Jesper Juhl


jjuhl:
I created the branch because I thought there would be a lot to fix by enabling various warnings, but in the end there was only this single trivial commit. Nice. You keep a clean code base :)

PR #280, opened 2026-01-02
Update copyright year to 2026

jjuhl:
It's now 2026, so the year in the copyright statement in various files should be updated from 2025 to 2026.


jjuhl:
Missed a few bits at first. Fixed that up.

PR #283, opened 2026-01-04
Include cleanup part1

jjuhl:
Hi Texus

Here's part one of the header cleanup we talked about on Discord (more will follow if you like/approve of this).

I'm splitting this into several parts since there's a lot to do and I don't want to spend the time doing everything at once if it turns out you don't like the end result. 
So, here's the first batch of changes for you to look at.
There's a single commit for each logical change, to make it easy to review.
Each include added is because the file it is added to directly uses something from the added header (files should include what they use).
Removed includes are removed because nothing in the file including the header uses anything from it.

If you like this, feel free to just merge it.
If you like it but want changes, please let me know and I'll update the PR.
If you hate it, please tell me why - then maybe I can adjust to your feedback or maybe just drop it entirely.

3 small things to be aware of:

 1) Your preferred include order is not observed by all files - I've ignored that for now and will do a commit at the very end of this that shuffles includes around to conform to your preference.

 2) Some files have a blank line between TGUI includes, dependency includes, std lib includes, but some do not. There doesn't seem to be a specific consistent style, so for now I've just ignored that detail and plan to just do a final pass over all files at the very end that either adds or removes those blank lines depending on what you prefer.
 
  3) Includes are not all currently sorted alphabetically and so I've not really cared about that. If you want them sorted then I'll also do that as a final pass towards the end so I don't have to think about it constantly while doing these patches.

Looking forward to your comments.

Kind regards 
Jesper Juhl


texus:
I still have very mixed feelings about this. The more I scroll through the diff file, the more I feel like this is a completely unnecessary change. If I'm honest, the biggest issue is that I personally just don't care about transitive includes at all. While I realize that the change might help avoid issues with refactoring, it is trivial and easy to fix the issues when they arise, so the additional effort up front and lots of extra lines of code aren't worth it to me. So even if I merge this change (to which I'm mostly indifferent, I'm not necessarily opposed to extra includes), I would not maintain the code to prevent new transitive includes.

For the std headers I mostly don't mind the changes, so my issues are mostly because I know that certain headers must include certain other headers to function so I find it wasteful to include them again. When I say something is "wasteful", I mean that it adds lines of code that are in my view unnecessary don't need to exist. They aren't bad to have per se, but I wouldn't write them.

I'll give some examples of cases that I immediately spotted when briefly scrolling through the diff file where I question the need. This whole discussion is more subjective than objective, so hopefully it gives more insight on my personal view on the matter. I apologize for the long and possibly incoherent rambling.

- TGUI/Config.hpp always needs to be included in every single TGUI header file. I feel like it's wasteful to repeat it everywhere when you know that any TGUI include will already transitively contain it. The only files that need to include it are those that don't include any other TGUI headers. In fact, I feel like it was already included in a few places too many.

- `<memory>` is needed by all widgets, it is inherent to them: you cannot create a widget without using `std::shared_ptr`. So I'm ok with it being included explicitly in Widget.hpp, but I find it wasteful to include it everywhere where a class derives from Widget (i.e. in any file inside the Widgets subdirectory).

- Color is to me inherently part of a renderer. Practically every renderer works with colors. So it seems wasteful to include them in every Renderer instead of just doing it explicitly in WidgetRenderer.hpp.

- In the .cpp files of the renderers I definitely don't like the Color.hpp includes, because you already know it is part of the interface and thus already included in the header. In my opinion the .hpp file only needs to include what is needed for the declaration and the .cpp file includes what is needed for the implementation. I don't like having to repeat everything in the .cpp file that was already needed for the declaration anyway. This is a very grey area because you might e.g. use `std::map` in a function implementation and somewhere inside the header file, and I'm fine with `<map>` being included again in the .cpp file in that case. However in case of the renderers I see no use why Color.hpp would be included in the .cpp files.

- For the tests I really liked that I could just include Tests.hpp and pull all needed dependencies (i.e. catch.hpp and TGUI.hpp). I feel like the tests should include TGUI the way I intend for people to use it, so they probably shouldn't contain a Color.hpp include in every test file (to be certain that my TGUI.hpp or Core.hpp interface actually contains everything that is needed in order to use TGUI). Right now Tests.hpp is including TGUI.hpp which pulls in everything. What I would be ok with is to include TGUI/Core.hpp instead and have each test include additional headers that aren't part of TGUI's core (e.g. there should be no include for Color.hpp but the tests for EditBox would need to include the EditBox.hpp header specifically). The Core.hpp file itself could be changed to reduce transitive includes (so an include to e.g. Color.hpp could be added inside Core.hpp to make it more clear that Color is part of the Core interface instead of having it silently be included transitively via Widget.hpp).

> Your preferred include order is not observed by all files

It's not a very strict rule, but I try to follow it when I'm paying attention to the order. Feel free to correct the order where it is different.

> Some files have a blank line between TGUI includes, dependency includes, std lib includes, but some do not.

There is no fixed style. If there are multiple includes in each group then I'm much more likely to add a newline compared to when there is only one file for each group. If I would have to standardize it then I would probably add the newline in between everywhere.

> Includes are not all currently sorted alphabetically

I don't care about the order in which they are listed. If they would be alphabetical everywhere already then I would probably keep using that order when adding new files. In a .cpp file, you should always include the .hpp file with the same name first, so that might cause issues with listing things alphabetically. Unless you always make that one .hpp include it's own group I guess.

jjuhl:
Thank you so much for your feedback.
I don't have much time right now or in the next few days to properly respond, but I'll review your comments properly in a day or two and provide a proper response.

Thank you for taking the time to review.

jjuhl:
OK, I've now had time to read your feedback properly and this is what I think I'll do moving forward:

First of all, I think we should just close this PR and then I'll just include patches on the few things we agree on along with other changes in the future.

As to the first part about transitive includes: I think we just have to agree to disagree. Personally I find it very nice to be able to easily see from the includes what bits and pieces a file uses. It's also very nice when refactoring (as you mentioned) that all files always include what they use, but I can also see the argument that it's a trivial fix when needed. Another reason I like explicit includes everywhere is that it usually makes it easier to move code to new compilers/standard libraries/operating systems, but again, it's not that hard to address when actually doing it.
So, we have different opinions on the matter and different habits, but that's fine - this is your project, so I'll just put aside my own preference and follow your lead/style - but thank you very much for taking the time to explain why you hold the views you do.

As far as the include order goes; I'll keep it in mind and fix up the ordering when I encounter something that doesn't follow it. I believe we can agree that that will be a good thing, for consistency :)

Regarding the blank lines between include groups; I personally would prefer no blank lines (but just enforce the order), but I'm fine with whatever and I guess at some point I may do a pass that just makes everything consistent (with the blank lines) but I may not - we'll see.

On the issue of implementation files including their own headers first I believe we agree, so I'll grab the one commit from this branch that fixes that up in a few places, maybe expand on it a bit, and add it to my next bunch of commits in a new PR.

As far as sorting commits alphabetically (within groups and with the primary header being an exception) ; that's just a habit I've developed over the years to be easily able to see at a glance if there are duplicate includes and to easily know where to add a new include and avoid adding a duplicate. If you don't care I won't bother with anything like that, but if you would like me to ensure all groups are sorted, just let me know and I won't mind doing it.

One thing you didn't touch on was the commit that removes includes that are clearly not used by the files including them. I hope we can agree that that makes sense and I'll include that commit (again; maybe expanding it or doing more such cleanups in seperate commits) in a future PR along with other fixes.

Does that make sense to you?

Kind regards
Jesper

texus:
Sound good.

I'm also ok with includes being in alphabetic order. So feel free to correct includes anywhere that you are making changes (or everywhere at once if you really want to).

Ideally something like clang-format would be configured for grouping and sorting the files. Then it wouldn't be just a matter of fixing them once, but then it could be enforced automatically in the future similar to what SFML has. I think I once started looking into it and gave up pretty quickly though.

> One thing you didn't touch on was the commit that removes includes that are clearly not used by the files including them. I hope we can agree that that makes sense and I'll include that commit

I forgot to check the individual commits, I was just scrolling through the unified diff. I definitely agree that it makes sense to keep that commit. And also the one that puts the own header first.

texus:
If you add those two commits ("Remove a few unused includes" and "Implementation files should include their own header first") to #285 then I'll just merge that PR afterwards.

jjuhl:
OK. I'll include the commits we agree on in a future PR and I may do a commit to sort includes.
I personally love clang-format and always use it for my own projects, so I'll take a look and see if I can come up with a rules file that matches your preference, then we could do a one-time cleanup and then a .clang-format file to enforce it - I'll look into that.

/Jesper

jjuhl:
Updated #285 with the commits we agreed on. :-)

PR #285, opened 2026-01-06
Mixed fixes

jjuhl:
Just a few small cleanups/fixes.


texus:
The `BackendFontFactory::createFont()`, `BackendRenderer::createText()` and `BackendRenderer::createTexture()` virtual functions can't be made const for backwards compatibility. They are base classes for custom backends. I've added a note to add the const if I ever start working on TGUI 2.

While the `contains()` function is better, that part of the code is a test for the `count()` function :smile:

jjuhl:
Makes perfect sense :)
I've dropped those two commits and rebased on the latest 1.x branch.

jjuhl:
Rescued a few commits from the include-cleanup-part1 branch/PR as we discussed.

PR #287, opened 2026-01-07
Fix a few cases of potentially incorrect rounding

jjuhl:
Adding +0.5 to a floating point value and then casting (truncating) to an integer is not the right way to do rounding. Although it may seem to do the right thing most of the time it has issues in several corner cases. The correct thing to do is to use
std::round/std::lround/std::llround that know how to do the rounding correctly including for the corner cases (and additionally, the compilers know those functions well and often generate more optimized code for them than random arithmetic).


jjuhl:
Some compilers on platforms I don't have access to are apparently not happy with this.. Close for now, I'll revise it.

PR #292, opened 2026-01-11
Don't use space btween if and condition in CMake files

jjuhl:
Looking at what is the common style in CMake files reveals that one should not have a space between the 'if' and the actual condition. That clashes very much with what I personally prefer, but maintaining concistency is much more important. So, here's a commit that makes the style consistent.


texus:
There are also CMakeLists.txt files in `tests/`, `src/Backend` and `src/` where similar changes can be made.

You accidentally changed a line in `examples/main-SFML_GRAPHICS.cpp`, which isn't a cmake file.

jjuhl:
Thank you for the feedback. I'll fix that up and update the PR.

PR #293, opened 2026-01-14
Miscellaneous

jjuhl:
Hi Texus

Here's another mixed bag of fixes/cleanups for you.
As always, let me know if you need me to make any changes, otherwise feel free to merge as-is :)

Kind regards,
Jesper Juhl


texus:
Thanks again for your work.

I cherry-picked those commits because I didn't like the ["Remove some instances of redundant member initialization" commit](https://github.com/texus/TGUI/pull/293/commits/6832ab83efe9a291f71715ecbf497ecd726f3026). I like having empty initializers in copy and move constructors to indicate that I'm doing something special. Otherwise it could raise a question whether I maybe forgot to add the member to the copy constructor (which is a mistake that has happened a few times already when adding new members). It also clearly shows for which members there is an exception, for when you wonder why the default copy constructor couldn't be used.

For the initializer list of normal constructors I would be ok with removing the redundant code, because not all members are listed there anyway (because most are being initialized at the declaration in the header file instead of in the initializer list).

Based on you making "Container::saveWidgetsToFile" const, I also made "BackendGui::saveWidgetsToFile" a const function. The BackendGui has almost the same interface as Container, but it doesn't inherit from it because a Gui isn't a widget and it would also inherit functionality that wouldn't make sense.

jjuhl:
Thank you for the feedback. I'll try to keep that in mind 🙂

PR #294, opened 2026-01-14
Misc fixes and cleanups

jjuhl:
Some cleanups for you - mostly just whitespace.


texus:
You probably shouldn't make changes to files in `extlibs/` and `examples/android/RAYLIB/app/src/main/cpp/deps/` because these are files from dependencies that may be copied verbatim from external projects. Similarly `tests/catch.hpp` should not be touched.
The changes to all other files are ok.

jjuhl:
👍 I'll fix that up soon (probably tomorrow).

jjuhl:
Fixed up according to your feedback and rebased on top of the current 1.x branch.

PR #297, opened 2026-01-24
Various stuff

jjuhl:
Hi Texus

Just a few small, relatively trivial, commits for you :)

/Jesper Juhl


texus:
Thanks. I don't like having the body on the same line as the if statement though, so for the fclose call I would prefer seeing something like this:
```c++
const auto closeFileFunc = [](FILE* fp){
    if (fclose(fp))
        throw tgui::Exception{U"Failed to fclose file, error: " + tgui::String(std::system_error(errno, std::generic_category()).what())};
};
```

PR #305, opened 2026-01-28
Small stuff

jjuhl:
Hi Texus

A small collection of small commits for you to consider :)

-- 
Jesper Juhl


jjuhl:
Rebased properly on top of 1.x

texus:
The std::move of trivial types is a personal choice. It doesn't happen in just that one file. I have wondered multiple times before if it wouldn't be better to remove the std::move, but I personally still find that the benefit outweights the downside.

The std::move indeed does absolutely nothing. However I like it in move constructors for the same reason that I like listing all members of the class there: to quickly tell why a custom move constructor has to exist. By removing the std::move in those lines, I now need to go check the class declaration to find out whether those are trivial types or whether I simply don't want their value erased in the moved-from object. With the std::move there, I can immediately tell that I don't care about any of the values in the moved-from object (except for m_messageTopicId which is set in the body).

To be clear: in any place other than the move constructor and move assignment operator I would be against the use of std::move on a trivial type. But I consider those functions to be an exception.

So unless you can manage to convince me to change my mind about this subject, I'd rather not merge the "Don't std::move trivially copyable types like std::uint64_t" commit.

jjuhl:
I know the std::move of trivial types happens in many places. This was just a tester to see if you agreed with me that we shouldn't do so - you don't agree, so I'll drop that commit and ignore that in the future. :)
We can always discuss later if it makes sense or not, but for now, let's just leave it alone.

PR #307, opened 2026-02-01
Various fixes

jjuhl:
Hi Texus

Here's another small batch of small fixes/cleanups for you to consider.

Kind regards
 Jesper Juhl


texus (commented):
The std::move in BackendGui::handleEvent isn't pointless here, is it? The local mouseMoveEvent variable isn't used anymore afterwards and the handleEvent function takes the event by value?

jjuhl:
I'm pretty sure the std::move is pointless.
Event doesn't have a move constructor nor a move assignment operator, so casting the passed in "mouseMoveEvent" to a xvalue via std::move can't really do anything. And the Event struct also looks like it's almost trivially copyable (to me - I could be wrong).

Also, if I compile the file with g++ and -O2 to like this:
g++ -c -m64 -O2 -std=c++17 -fPIC -fvisibility=hidden -fvisibility-inlines-hidden -fdiagnostics-color=always -Wall -Wextra -Wshadow -Wsign-conversion -Wfloat-conversion -Wnon-virtual-dtor -Wold-style-cast -Wcast-align -Wunused -Woverloaded-virtual -Wpedantic -Wdouble-promotion -Wformat=2 -Wimplicit-fallthrough -Wnull-dereference -Wmisleading-indentation -Wduplicated-cond -Wlogical-op -Wsuggest-override -Winvalid-pch -I/home/jj/src/TGUI/include -DTGUI_DYNAMIC -DTGUI_EXPORTS -x c++ /home/jj/src/TGUI/src/Backend/Window/BackendGui.cpp
and then compare the result before and after the change I get this:

before:
  $ size ./BackendGui.o
   text    data     bss     dec     hex filename
  21368     336       0   21704    54c8 ./BackendGui.o
after:
  $ size ./BackendGui.o
   text    data     bss     dec     hex filename
  21368     336       0   21704    54c8 ./BackendGui.o

and if I then do 
  $ objdump -C -d ./BackendGui.o
the generated assembly for BackendGui::handleEvent is also identical in both cases.

texus:
You are absolutely right. I was only looking at the calling code and at the function signature, I didn't think about the fact that Event was a trivial type.

PR #311, opened 2026-02-07
Misc small cleanups

jjuhl:
Hi Texus

Here's another small batch of small cleanups for you to consider :-)

Kind regards
Jesper Juhl


texus:
Some feedback on 2 of your commits. I will cherry-pick the others and close this PR in a moment.


### Avoid global names in header files

The global `tgui::priv::nsvg...` names have a reason to exist. When TGUI_USE_SYSTEM_NANOSVG is 1, a nanosvg.h file that isn't bundled with TGUI will be included. This is a C header, so it's functions are global. Your changes in src/SvgImage.cpp wouldn't compile when TGUI_USE_SYSTEM_NANOSVG is set because `tgui::priv::` does not exist in that case.
The namespace thing is bit of a hack to avoid duplicate symbols in case TGUI uses it's bundled nanosvg.h and your own project would also use that library.

Note that the `TGUI/extlibs/Include...` header files are helper files to include dependencies. They exist so that a bunch of preprocessor code doesn't need to exist in every source file that needs that dependency. They are not meant to be included by the user, so they usually don't polute much. In this particular case, `IncludeNanoSVG.hpp` is only included in `src/SvgImage.cpp`.


### Simplify some code by using std::any_of/std::all_of

I'm not used to using these functions, so I personally find the code slightly harder to read. That's just lack of experience with them though, I do believe it is good to use them and I should probably learn to use them more myself.

However, (and maybe it's just because I'm not used to them), I feel like it's wrong to use these functions when the body has side effects. I know that these functions will still execute the body sequentially, but I feel like any_of and all_of should only be used if the code COULD be parallelized in the future. So I don't really like the `any_of` call in `TreeView.cpp` because the call to markNodesDirty is a side-effect.

I have less of a problem with `all_of` in `tests/Focus.cpp`: the lambda only returns true or false without doing anything else.
I still have a small personal preference for a normal loop here because I'm not used to code like that were the lambda is more than one line, but I wouldn't mind if you changed more code to be like that.

I'm not merging the commit now because of the `any_of` call in TreeView.cpp, but if a future PR contains the `all_of` call in tests/Focus.cpp then I would definitely merge it.

jjuhl:
Thanks a lot for your feedback. I''ll include the all_of change in a future PR and I'll keep the other feedback in mind.

PR #314, opened 2026-02-08
Various small fixups

jjuhl:
Yet another small batch of small fixups for you to consider :)


texus:
The lines on top of the HTML file are special. It is indeed invalid, but the page is still displayed by the browser and it was the most ~~genius~~ lazy option. I'll explain the issue below, I'm open for alternatives as it doesn't cost me too much extra work every time I want to make a new TGUI release.

The documentation needs to be generated for 2 use cases:
- For anyone who want to generate the documentation on their pc (I'm not sure how many do this though).
- For the website.

For the website the documentation needs additional header and footer code to fit in. The navigation bar from the website should still be there for example, so that you can go back to other parts of the website while you visit the online documentation. The documentation generated locally can't use the website layout though. So there would need to exist 2 separate versions of the documentation that aren't compatible. Or so you would think...

Every page on the website is processed by Jekyll if it starts with a YAML header. An empty YAML header consists of 2 lines of "---", which is what you found in doc/header.html. The header file contains some really hacky lines such that it will include other files when parsed by Jekyll, but has all that code inside HTML comments when opening the .html file directly. This allows me to use the exact same documentation files for both use cases. When I need to update the website, I literally copy the locally generated files without any modifications.

The downside to this solution was that `--- ---` is displayed at the top of the page when viewing it locally, and that it is indeed invalid HTML.

jjuhl:
Thank you for explaining.
I'll update the PR later and drop that commit.
I'll have a think about a better solution to the HTML issue and if I come up with something I'll include it in a future PR.

jjuhl:
PR updated with the HTML commit dropped.

PR #315, opened 2026-02-11
Minor cleanups

jjuhl:
Hi Texus

Yet another batch of tiny cleanups for you :)


texus:
I'm a bit worried about the includes being removed from src/Filesystem.cpp, even though CI seems happy about it.
I placed a comment next to those headers to indicate for which function they were being included. Both `stat` and `errno` are still used in that source file even though the includes are now gone. So either it gets `stat` and `errno` from some other includes, or that specific code isn't being compiled in any of the tests (since it is platform-specific and depends on another define).

jjuhl:
You make some good points. I've removed that bit of the commit.

PR #320, opened 2026-02-19
Mixed commits

jjuhl:
Just a few small changes.


texus:
I don't think the widening casts are needed. They just make the code harder to read.
If we would have to start worrying about containers with a billion elements then we would also need to check that every addition doesn't cause an overflow. At some point it just becomes too theoretical to bother with it.

jjuhl:
Fair comment. I've dropped the "widening casts" commit.

PR #322, opened 2026-02-21
Add a few test cases for simple situations

jjuhl:
Hi Texus

This is just a "feeler PR" to find out if you will appreciate additions to the test suite.
These tests are fairly basic (but relevant IMHO).
I'll add more in the future if you like them - just don't want to waste a bunch of time writing tests if you don't want them :)

### DIFF
--- a/tests/Widgets/FileDialog.cpp
+++ b/tests/Widgets/FileDialog.cpp
@@ -96,11 +96,13 @@ TEST_CASE("[FileDialog]")
 
     SECTION("Captions")
     {
+        dialog->setCreateFolderButtonText("Summon folder");
         dialog->setConfirmButtonText("Use that file");
         dialog->setCancelButtonText("Don't do it!");
         dialog->setFilenameLabelText("What file?");
         dialog->setListViewColumnCaptions("Moniker", "Largeness", "Timing");
 
+        REQUIRE(dialog->getCreateFolderButtonText() == "Summon folder");
         REQUIRE(dialog->getConfirmButtonText() == "Use that file");
         REQUIRE(dialog->getCancelButtonText() == "Don't do it!");
         REQUIRE(dialog->getFilenameLabelText() == "What file?");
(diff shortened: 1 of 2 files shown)


texus:
More tests are welcome, especially if they touch on situations that aren't already covered.
If some lines of code are only tested indirectly by internal function calls then I still think it is good to have explicit tests for them. However if other test cases already call certain functions then I don't necessarily feel like new tests focusing on just these functions are very useful.

I would merge the "Child Widgets" tests because more explicit tests are never bad, but they are unlikely to add much value because this is already covered by the Container tests. They might protect against a future case where Panel would override the add or remove functions, but if tests are done properly then new tests would be written together with such a change to Panel anyway.

Time might be better spend writing tests for parts of the code that are untested or partially untested. You can use https://app.codecov.io/github/texus/TGUI to see which part of the code is covered by the tests.

The addition in SignalManager.cpp doesn't seem to have anything to do with SignalManager (which is an actual class for an edge-case situation, not the place to tests signals). The signal class itself (so e.g. `connect` and `emit` functions) is tested in Signal.cpp, and each widget has tests for its own signals. So tests/Widgets/Button.cpp has tests for connecting to the Pressed signal in it's "Signals" section and also has tests to verify that pressing it triggers it in the "Events / Signals" section. Many other widgets don't have a complete "Events / Signals" section though, that is e.g. something that could still use some work.

jjuhl:
Thank you for the feedback.
I've dropped the SignalManager bits.
The https://app.codecov.io/github/texus/TGUI link is super useful (didn't know about that one) - I'll take a look and see where there's potential to add more tests for currently untested code that can have a real impact. :-)

jjuhl:
Added a test for FileDialog::[get/set]CreateFolderButtonText that didn't previously have any test coverage.

I won't touch this branch any further, but will add more tests in a future PR if you are happy with this stuff :)

PR #328, opened 2026-03-02
Random changes

jjuhl:
Just a few random things I've noticed while browsing through the code.


texus:
The failed android build is because the Base64 header still needs to include `<TGUI/Config.hpp>`.
The rest looks good.

jjuhl:
> The failed android build is because the Base64 header still needs to include `<TGUI/Config.hpp>`. The rest looks good.

Yeah, I tried that and it also needed StringView.hpp, but even after adding those I still got failures, so in the end I just dropped all the changes to the files that caused issues - it can always be revisited some other time :)

PR #332, opened 2026-03-06
Various mixed commits

jjuhl:
A mixed bag of stuff for you :)

### DIFF
--- a/gui-builder/src/main.cpp
+++ b/gui-builder/src/main.cpp
@@ -67,7 +67,7 @@ int main(int, char* argv[])
         if (argvW == nullptr)
         {
             std::cerr << "Failed to access command line arguments\n";
-            return 1;
+            return EXIT_FAILURE;
         }
 
         const tgui::String exePath = argvW[0];
--- a/include/TGUI/Global.hpp
+++ b/include/TGUI/Global.hpp
@@ -63,6 +63,7 @@ namespace tgui
     template <typename T>
     TGUI_NODISCARD constexpr const T& clamp(const T& v, const T& lo, const T& hi)
     {
+        TGUI_ASSERT(!(hi < lo), "The highest value must not be less than the lowest value");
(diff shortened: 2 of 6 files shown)


jjuhl:
Hmm, the failure in CI (linux-latest-dev) doesn't happen on my local machine. I'll just leave the PR as-is and maybe you can figure out what the issue is :)

texus:
> maybe you can figure out what the issue is :)

I checked other places where `setEditCursorBlinkRate` was used and found that is was being changed at https://github.com/texus/TGUI/blob/6759d477281fc03c18b4e14f9d5f720519fad486/tests/Widgets/EditBox.cpp#L356

Unlike your code, my test didn't restore the original value when it's done. So if that test runs before yours then it will fail.
While you could technically fix the test in EditBox.cpp, it is actually the same as what you are testing, so I think you can just remove the entire "Blink rate" section in EditBox.cpp and keep the new one in Global.cpp.

> TGUI_ASSERT(!(hi < lo), "The highest value must not be less than the lowest value");

Maybe `!(hi < lo)` can be replaced by `hi >= lo`?
Both are valid, with the negation it might read more like "you can't do this thing" as opposed to "you must do this thing", but I find it a bit weird to see a negated comparison.
This is nitpicking of course, if I wouldn't have already had to comment on setEditCursorBlinkRate then I wouldn't have bothered mentioning it and would have just merged it.

jjuhl:
"Maybe !(hi < lo) can be replaced by hi >= lo?" - I considered that but rejected it since most types implement operator< but not necessarily other operators, so this is the form that would work for most types.

jjuhl:
I'll update the PR regarding blink rate.

texus:
I don't think the clamp function will ever be used on a non-numeric type that lacks a `>=` operator, but it is a valid point.
Using `!(hi < lo)` would indeed be better for not putting additional restrictions on the template type.

PR #345, opened 2026-04-03
A few cleanups

jjuhl:


texus:
Putting the macro parameter in brackets feels unnecessary to me because you can't really use an expression there (as the parameter is used as the name of a variable), but I'm still accepting it because it is good practice in macros in general.

Issue #325, opened 2026-02-28
Support for themes in forms


jpgaribotti:
Widgets created through `loadWidgetsFromFile` can only have renderers defined in the file itself. Even if you import a theme in gui-builder, it only copies the theme into the form. If a default theme is set, they won't pick up the default theme, and won't support "hot reloading" if you load a different theme like programatically created widgets do. This makes development somewhat less convenient, and requires duplicating the theme in every form and keeping them up to date as they are modified.

It would be nice to be able to pass a theme to loadWidgetsFromFile and have the widgets either pick the renderer for their type from there, or be able to reference (with the & syntax or a specific syntax) renderers from the theme.

I currently use a workaround for this that after loading from file walks through every widget and sets renderers (I store theme renderer names in UserData if I want something other than the default), but this is not ideal.

I could send a PR if this is something you'd be interested in supporting.

texus:
One of the issues of storing a theme in a form file is that the location of the theme file will be different for the Gui Builder and for your application.

The form definitely needs some annotation for whether to load from the runtime theme or from the custom hardcoded renderers inside the form, to make certain that you can support e.g. labels of multiple colors. Ideally it should still be possible to use multiple themes in a single form file. This however quickly leads to further issues like what to do when using multiple themes that have the same name but are from different folders.

Suppose you have a form like this:
```
Renderer.1 {
}
Renderer.2 {
}

Button.Button1 {
    Renderer = &1;
}
Button.Button2 {
    Renderer = &1;
}
EditBox.EditBox1 {
    Renderer = &2;
}
EditBox.EditBox2 {
    Renderer = {
    }
}
```

Maybe a new section could be added that looks like this: (this is just a quick idea, I don't know how easy it would be to adjust and maybe there are better alternatives)
```
Theme.White
{
    Button = &1;
    EditBox = &2;
}
```

Then loadWidgetsFromFile could optionally replace the renderers with a runtime theme when a map like `{"White": runtimeTheme1, "Black": runtimeTheme2}` is passed. In this example EditBox2 would keep it's custom theme since it isn't linked with the theme. The form is still self-contained like this, if no themes are passed to loadWidgetsFromFile then the form still has renderers for all widgets.

So there are multiple issues that need to be sorted out in order to properly support themes in form files. If you have a proposed solution then I'll definitely be interested in it.

jpgaribotti:
[...]

PR #348, opened 2026-04-05
Add themes to form loading (Issue #325)

jpgaribotti:
This PR has been broken into individual commits that can be built and tested separately for ease of review. Some commits only add interface changes and delay implementation to a later commit.

### Summary

This change implements **theme-aware form loading** so `.txt` forms can bind renderers to named themes and optional in-form fallbacks, with explicit load options on `Container` / `BackendGui`. It closes the gap between file and stream loading for default-theme handling and adds tests around the new behavior.

Proposed in [Issue #325](https://github.com/texus/TGUI/issues/325).

### Theme / form loading

- **`FormLoadOptions`** (`themesByAlias`, `applyDefaultTheme`) is accepted by overloads of `loadWidgetsFromFile` and `loadWidgetsFromStream` (lvalue and rvalue stream).
- **`loadWidgetsFromStream`** now follows the same default-theme clear/restore pattern as **`loadWidgetsFromFile`**, unless `applyDefaultTheme` is used to keep the global default theme during load (so widgets without a `Renderer` line match programmatic construction when desired).
- Form files may define top-level **`Theme.<Alias>`** sections with per-widget renderer fallbacks (e.g. `Button = &1;`).
- Widget **`Renderer`** values may use **`@Alias`** / **`@Alias.Section`**, resolved from **`FormLoadOptions::themesByAlias`** at runtime and/or form-defined fallbacks.
- Widget loading now takes **`WidgetLoadResources`** (extends the previous renderers map with theme alias resolution), with **`Container::loadWidgetsFromNodeTree`** available where a parsed node tree is loaded with options.

### Tests

- Added/extended loading tests (e.g. `tests/Loading/FormThemeLoad.cpp`) covering alias binding, `applyDefaultTheme`, and parity with the two-argument stream API when using default `FormLoadOptions`.

### Backward compatibility

Existing two-argument `loadWidgetsFromFile` / `loadWidgetsFromStream` call sites keep prior behavior by routing through default-constructed **`FormLoadOptions`**.


texus:
I haven't had the time to look at all the changes in detail yet, but it looked fine on first sight except for one big thing.

Changes to the Widget::load function signature break the backwards-compatibility of the API, as anyone with a custom widget will need to make this change. Code that was written to use an earlier TGUI 1.x version should ideally continue to work without changes.

Annoyingly, this means you also can't simply add another load overload and call that one, because then the old function would no longer be called. The old overload must continue to function correctly. One way to do it might be something like this:
```c++
void Widget::load(const std::unique_ptr<DataIO::Node>& node, const LoadingRenderersMap& renderers)
{
    // Implementation here. Does same as before if m_loadRuntimeThemesByAlias and m_loadThemeFallbacks are nullptr.
    
    // Would contains something like this:
    //   WidgetLoadResources widgetResources;
    //   widgetResources.renderers = availableRenderers;
    //   widgetResources.runtimeThemesByAlias = m_loadRuntimeThemesByAlias;
    //   widgetResources.themeFallbacks = m_loadThemeFallbacks;
    //   loadRendererFromFormValue(this, node->propertyValuePairs[U"Renderer"]->value, widgetResources);

    // Implementation of Container::load has to be careful with how it calls the Widget::load function on itself.
}

// This function overload would NOT be virtual
void Widget::load(const std::unique_ptr<DataIO::Node>& node, const WidgetLoadResources& resources)
{
    m_loadRuntimeThemesByAlias = resources.runtimeThemesByAlias;
    m_loadThemeFallbacks = resources.themeFallbacks;
    load(renderers);
    m_loadRuntimeThemesByAlias = nullptr;
    m_loadThemeFallbacks = nullptr;
}
```

If I ever make breaking changes in the future then I will replace the code to implement it like you did right now.

jpgaribotti:
Ah, I see what you mean, I kept backwards compatibility of the loadWidgetsFrom... functions and forgot the existence of client subclasses. I got too focused on my feature and didn't respect the API on a minor version bump.

We can't have an overload because if we call `widgetPtr->load(node, resources)` on a subclass that overrode `load(node, renderers)`, it will not run the subclass' custom code, leading to a broken widget. So we have to always call `load(node, renderers)`, assuming the old behavior, and opt-in to the new behavior by setting the extra variables on a side-channel instead of passing them as a parameter.

I'll have a look at it and update the PR to maintain backwards compatibility of the Widget API.

jpgaribotti:
@texus updated the PR to avoid breaking the API, following your suggested approach. Added various tests to confirm containers and pre-existing subclasses continue to work as expected. Rebased the changes to avoid having the API break in the history.

texus:
Thanks already for your work so far, I'll try to properly review this when I find enough time.

One thing that I would like to know is what happens if the default theme uses textures, the renderer in the form only uses colors, and you set `applyDefaultTheme` to true? I'm worried that it might still use the textures from the default theme as the renderer of the form might not touch the texture property (which could have been the reason why I previously explicitly set the default theme to nullptr).

Other than the above concern that simply needs to be tested, scrolling through the changes only resulted in some unimportant nitpicks:
- The loadWidgetsFromNodeTree declaration could use a brief description in header file.
- Maybe documentation could be added to the fields in FormLoadOptions?
- The changelog should refer to `PR #348` instead of `Issue #325`. When something in the changelog is contributed by someone then I usually add a PR link as a way to provide credit. There is a chance that I'll modify the changelog text after merging this, because usually it only provides a high-level overview of the change without providing details. However I'm not sure where else I would put the details in this case, so I'll have to think about that when I'm reading it more carefully later.

jpgaribotti:
Regarding your question, there's a main answer and some somewhat surprising side-effects but they don't alter the key point. I've added a test in the latest commit that highlights the relevant aspects of the answer below, so that there's no ambiguity as to what I mean.

**The default theme’s texture does not remain on the widget as the same loaded image** when the form replaces the renderer with a sparse inline `RendererData` (only `TextColor`, etc.). The inline data is still built only from the form; `RendererData::createFromDataIONode` does not merge the global default theme into that map.

Despite the widget not having a texture, **you can still see a `Texture` entry on the widget after load** when `applyDefaultTheme == true` and the *previous* renderer (from construction) contained a `Texture` property. That entry is typically an **empty** `Texture` inserted as a **side effect of a `const` getter**, not the theme’s PNG.

**`applyDefaultTheme` true vs false is not guaranteed to produce identical `propertyValuePairs` maps** for the same form, for the same reason: `Widget::setRenderer` notifies the widget about properties present on the *old* renderer but missing on the new one. If the old renderer had `Texture`, `ButtonBase::rendererChanged("Texture")` runs and calls `ButtonRenderer::getTexture()`, which **mutates** the map when the key is absent (adds an empty Texture entry).

So the case that concerns you, as I understand it, does not functionally happen. The widget's renderer is left with a Texture entry, but it's empty, as oppossed to the entry not existing in the map in the case of `applyDefaultTheme = false`.

I've also addressed the other points in a second commit. Regarding the changelog, I added it mainly as a reference, I can remove it if it makes things easier.

texus:
I've pushed a change that simplified some code, you may want to check if everything still works for you before I merge this PR.

PR #350, opened 2026-04-09
A random collection of commits

jjuhl:


texus:
I'm a bit concerned about that a `filesystem-test-dir` folder would exist in the root directory.
Are you running the tests from the root directory? The tests also write a lot of .png and .txt files (using `tgui::writeFile`) which I would expect are created to the same directory as where `filesystem-test-dir` is created. So either a lot more files end up in your root directory, or somehow there is a difference in the working directory between the `tgui::writeFile` and `tgui::Filesystem::createDirectory` calls.

So while I don't mind adding `/filesystem-test-dir` to .gitignore, I'd like to first understand why it exists in the first place.

jjuhl:
I usually run tests like this :

> [jj@smurf TGUI]$ pwd
> /home/jj/src/TGUI
> [jj@smurf TGUI]$ (cd build/Desktop-Debug/tests/ ; ./tests)
> ===============================================================================
> All tests passed (11902 assertions in 81 test cases)
> 
> [jj@smurf TGUI]$ 
> 

I have tried running tests from qtcreator a few times and that might be what causes it, I'm not sure.
I currently have no idea why exactly that directory ended up in the root of my project directory, but it's there:

> [jj@smurf TGUI]$ ls -ld filesystem-test-dir/
> drwxr-xr-x 2 jj jj 4096  1 jan 20:58 filesystem-test-dir/
> 

I'll try deleting it and run some tests to see if I can discover the reason. I'll get back to you with what I discover (if anything).

jjuhl:
Ok, so I've spent some time digging around and experimenting and the only way I can find to get the filesystem-test-dir to show up in the repository root is to actually attempt to run the tests from there (which obviously fails and leaves the directory behind). I guess I must have done that by accident without realizing.  So, I guess we can just drop the commit that adds it to .gitignore or keep it if you like - your choice.

texus:
Instead of ignoring the `filesystem-test-dir` directory, I've updated the code to prevent the tests from running when the resources aren't found (441131151dff6f0b40267f835fdd88a7ca3442e2). This way it will never create the `filesystem-test-dir` directory if you accidentally run it from the root folder, so nothing needs to be ignored.
I've merged the other commits.

PR #352, opened 2026-04-13
Add more tests

jjuhl:


jjuhl:
Hmm, That failure is surprising to me. 1) The test succeeds on my Arch Linux machine. 2) When reading the code I very much expect an exception to be thrown.

texus:
CI is running the tests with multiple backends.
Based on the output, the 5th run, which uses the RAYLIB backend is the one that fails. So that backend probably does something differently. I can look into this later if needed.

jjuhl:
I don't (currently) have a way to test with that backend, so if you can check what's going on there I'd appreciate it :-)

texus:
The [BackendFontRaylib::loadFromMemory](https://github.com/texus/TGUI/blob/1.x/src/Backend/Font/Raylib/BackendFontRaylib.cpp#L39) function doesn't actually load the font. It always returns true no matter what you pass to it. The font data is only accessed once it needs to read a character from it.

If you want to guarantee that `tgui::Font(nullptr, 0)` always throws when data is a nullptr then you can replace `return true;` with `return m_fileContents != nullptr` in [BackendFontRaylib::loadFromMemory](https://github.com/texus/TGUI/blob/1.x/src/Backend/Font/Raylib/BackendFontRaylib.cpp#L72).

jjuhl:
I'll look into that :)

texus:
Interesting, because `new std::uint8_t[0]` isn't a nullptr, m_fileContents is actually a valid pointer in BackendFontRaylib::loadFromMemory. So maybe it would be better to check if `data` is a nullptr in [BackendFont::loadFromMemory](https://github.com/texus/TGUI/blob/2047c3416470a92884175e5005554bba76de2a20/src/Backend/Font/BackendFont.cpp#L110) instead.

texus:
I've patched BackendFont::loadFromMemory and cherry-picked the changes from this PR.

PR #356, opened 2026-05-01
Misc minor cleanups

jjuhl:


texus:
I don't really like the "Simplify tgui::utf::encodeCharUtf8 slightly" change. I generally prefer to have the code in the `else` branch, unless the else branch is the one that is usually taken.
If most of the time you needed 4 bytes then the change would make sense to me, but since it can randomly be 2, 3 or 4, I feel like the variables shouldn't be initialized to a value that is often going to be overwritten a moment later.

jjuhl:
Fair enough. I don't necessarily agree 100%, but I can respect your point of view, I'll drop that commit.

jjuhl:
Hmm, Your CI seems to have issues. All I did was drop the commit in question and reword another one.

PR #359, opened 2026-05-08
A few more tests

jjuhl:
Just a few basic tests.


texus:
For endsWithIgnoreCase and startsWithIgnoreCase I would also verify that "Hello world" starts/ends with "Hello world"

PR #361, opened 2026-05-11
Initial c++17 cleanups

jjuhl:
Hi Texus

Here are a few initial C++17 cleanup patches for you.
There's much more in the pipeline but I don't have time to do more than this tonight, so I thought I'd just submit what I had done so far - more will come in future PR's.


jjuhl:
Hmm, this apparently fails while building a C++20 module. I must admit that that is not something I have any experience with.. If you have an easy fix that would be great, otherwise I'll dig into it over the next couple of days.

texus:
For the c++20 modules, all classes and global functions must be explicitly marked for export in `src/TGUI.cppm`. The `tgui::clamp` function is listed there, but it couldn't be found anymore because it was removed.
I had already forgotten that the TGUI.cppm file existed, it's likely some new things got added which I forgot to add there. At least when you remove something you get a compile error to tell you that you forgot (although an error in that file also means you are breaking the public API).

Even though TGUI code should no longer use TGUI_NODISCARD or tgui::clamp anywhere, custom widgets in user code might still be using them. So I'm keeping them available to keep backwards compatibility.

PR #365, opened 2026-05-15
Misc c++17 and other stuff

jjuhl:


jjuhl:
:-( seems my environment is too modern and I cannot make the C++17 assumptions I did everywhere.

texus:
`starts_with` and `ends_with` were only added in C++20

(201703L is C++17, so 201711L is newer, I'm assuming you saw a check against the "2017xx" value and assumed it was checking for C++17)

jjuhl:
I'll get it fixed up. Thanks for the comment.

PR #368, opened 2026-05-20
Some small stuff

jjuhl:


texus:
I'm a bit torn about the "Remove some redundant access specifiers" commit.

While it's technically correct, I've always preferred to separate functions and members and I've abused the access specifiers a bit to be the title for each group (due to it being on a different indentation). Ignoring "protected" for simplicity here, you would have 4 groups inside a class:
- public functions
- private functions
- public variables
- private variables

Each group would start with an access specifier. Since public variables often doesn't exist, it did lead to "private" being repeated for the functions and variables section.

So while I'm not heavily against the change, I still have a preference for the duplicate access specifier as small extra indication that you switch from functions to variables.

jjuhl:
I don't feel strongly about it. It's just something I've never done myself and I noticed it, so I cleaned it up.
Feel free to keep or drop the commit as you see fit and I'll go with whatever you choose in the future.

texus:
Then at least for now it would be best if you just dropped that commit.

jjuhl:
I'll drop it and update the PR tomorrow. 🙂

PR #375, opened 2026-05-29
Some random patches

jjuhl:


texus:
> Replace std::list with std::vector in Layout constructor

On first sight it does look like you are correct and std::vector can be used here as the `erase` calls seem to happen on elements BEHIND the iterator that is kept. However maybe you can also get rid of the `std::advance` usages?

For example
```c++
auto nextOperandIt = operandIt;
std::advance(nextOperandIt, 1);
```
looks like it could just be the following now:
```c++
const auto nextOperandIt = operandIt + 1;
```

`std::advance` was only used because `std::list` was used.

> Simplify CSS

The `max-width: 100%;` probably needs to stay. This custom.css file exists to make the generated code compatible with the website. And on the website the `max-width` of `body` is altered (https://github.com/texus/TGUI-website/blob/master/resources/style.scss#L22).

jjuhl:
Ok. I'll drop the max-width change.
As for the std::advance, I did think of that but chose to maybe make that a future commit since I couldn't convince myself that changing it would be 100% safe, so I just stuck to what I knew would be safe (and then more changes can maybe come later).

jjuhl:
Dropped the max-width change.
The other bits can maybe come in future commits. But thank you very much for pointing it out :-)

PR #379, opened 2026-06-03
Msvc standard compliance

jjuhl:


jjuhl:
Ok, so it seems clang-cl doesn't understand /Zc:enumTypes - should I just drop that last commit or special case msvc vs clang-cl? Or just drop this entire branch? The intention here is to get msvc to behave as standard compliant as possible so we can in general just go by what the language standard specifies and worry less about compiler differences.

texus:
If I understand what `/Zc:enumTypes` does correctly, then I'm not sure if it is a good idea for enabling it when building TGUI even with MSVC. It feels like this is an option that needs to be set for all dependencies in your project as well, otherwise you risk binary incompatibility (e.g. the implementation and calling code use a different size for the function parameter).

The other options don't cause such potential issues so they are fine.
Technically the `/Zc:__cplusplus` option also alters how the header looks like when compiling TGUI vs when compiling the final project, but the headers should never use `__cplusplus` in cases where this would cause an issue (the TGUI_COMPILED_WITH_CPP_VER define exists for these cases).

jjuhl:
Right, good point. I dropped that commit.

PR #380, opened 2026-06-05
Various random improvements

jjuhl:


texus:
As far as I can tell, `volatile` isn't used anywhere. And I don't think I will ever use it in this project. So I don't really think this is needed.

jjuhl:
I also don't think it is currently used. I just wanted to add the flag as a "make sure MSVC behaves as much as the other compilers and as standard compliant as possible" safeguard.
Close or merge the PR as you see fit :)

PR #390, opened 2026-08-01
Use .contains() when possible

jjuhl:
Much more readable.


texus:
That's a C++20 feature

jjuhl:
Whoops. Sorry about that I forgot build as C++17, so I didn't notice :(

PR #392, opened 2026-08-09
Avoid an unneeded copy in TreeView::changeItemHierarchy

jjuhl:
We don't need to copy-construct 'nodeSharedPtr', a const reference is fine for what we do with it, so just hold onto a const& instead of creating a copy.


jjuhl:
Whoops. This is wrong. It worked on my machine, but fails CI, so obviously I've missed something.

PR #396, opened 2026-09-12
Move a few functions into anonymous namespace

jjuhl:


texus:
I don't really think it's needed for such a simple example code.
If it weren't for the multiple backends, then that entire example would essentially consist of a single source file.

jjuhl:
Fair enough :)

